PR #31, opened 2020-01-04
registration fix

ivava777:


src/main/js/src/app/_services/auth.service.ts
```diff
@@ -44,79 +45,87 @@ export class AuthService {
...
-        });
-
+        if (this.useServer) {
+            this.curLogin.usernameOrEmail = username;
+            this.curLogin.password = password;
+            this._http.post(this.urlLogin + '/auth/login', this.curLogin).toPromise()
+            .then(data => {
+                // const currentUser: User = this.getUserFromData(data);
+                const currentUser: User = {///TODO check data from server
+                    id: 0,
```
rpkyrych: You don't have to send id on back-side. We don't need it

PR #46, opened 2020-02-02
new files

Wolshebnik:
Front-End
Create new Registration form


src/main/resources/static/js/popup/ConnectBase.js
```diff
@@ -0,0 +1,23 @@
+class ConnectBase {
+  constructor() {
+    this._url = "http://localhost:3000/auth/";
```
rpkyrych: Java server runs on port 8080, so if you're going to send request to back-end you have to change port.

src/main/resources/static/js/popup/ConnectBase.js
```diff
@@ -0,0 +1,23 @@
...
+    this.headers = {
+      "Content-Type": "application/json"
```
rpkyrych: Should we compile you code or it gonna work without it? 
If it needs to be compiled so it becomes additional server and you have to put additional headers for CORS requests.
If not - everything will work fine.

src/main/resources/static/js/popup/ConnectBase.js
```diff
@@ -0,0 +1,23 @@
...
+    };
+  }
+
+  async postRequest(body, value) {
+    let url;
```
rpkyrych: You may initialize this variable by `this._url` and add "login" or "register" in line 12 and 14 as well.

src/main/resources/static/js/popup/ConnectBase.js
```diff
@@ -0,0 +1,23 @@
...
+      url = this._url + "login";
+    } else {
+      url = this._url + "register";
+      document.querySelector(".auth").style.display = "none";
+    }
+    return await fetch(url, {
+      method: "POST",
+      headers: this.headers,
+      body: JSON.stringify(body)
+    }).then(data => data.json());
```
rpkyrych: You should add exception handling also or make few attempts for sending the request. We have pretty dumb server so I couldn't guarantee that your request will be send from first attempt.
Also from back side we response with different codes and bodies. Try it with swagger-ui: http://54.37.125.180:8080/swagger-ui.html#/

src/main/resources/static/js/popup/validation.js
```diff
@@ -0,0 +1,136 @@
...
+      return rulesPattern[pattern].test(el.value);
+    },
+    contains(el, el2) {
+      if (el.value === formField[el2].value) {
+        formField[el2].nextElementSibling.innerHTML = "";
+        return true;
+      } else return false;
+    },
+    notMore(el) {
+      return el.value.length < 15;
```
rpkyrych: Why did you hard-coded 15 here? Please move it to constant or variable with name that will describe it's purpose.

src/main/resources/static/js/popup/validation.js
```diff
@@ -0,0 +1,136 @@
...
+    },
+    lengthLogin(el) {
+      return el.value.length > 2 && !el.value.includes(" ");
```
rpkyrych: Another 'magic number'?

PR #55, opened 2020-02-23
CGE-62

21November:
Added redirect to index page when root address is entered.


src/main/java/com/hillel/items_exchange/config/MvcConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.core.Ordered;
+import org.springframework.web.servlet.config.annotation.*;
+import org.springframework.web.servlet.view.JstlView;
+import org.springframework.web.servlet.view.UrlBasedViewResolver;
+
+
+@Configuration
+@EnableWebMvc
+@ComponentScan("com")
+public class MvcConfiguration implements WebMvcConfigurer{
```
rpkyrych: This class is excess - you should completely remove it.

src/main/resources/application.properties
```diff
@@ -14,15 +14,16 @@ spring.datasource.password=
...
-spring.liquibase.user=root
-spring.liquibase.password=
 
 ### Hibernate Properties
 # The SQL dialect makes Hibernate generate better SQL for the chosen database
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL8Dialect
 spring.jpa.hibernate.ddl-auto=update
 
+## Spring View Resolver
+spring.mvc.view.prefix=html/
```
rpkyrych: I don't see the same configuration for test purpuses.

21November: Performed.

src/main/java/com/hillel/items_exchange/config/MvcConfiguration.java
21November: Performed.

PR #60, opened 2020-03-04
CGE-51 Create REST service for getting the item's images

lubchik-14:
Create REST controller for getting all images by item_id.


src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+
+@RestController
+@RequestMapping("/products")
```
rpkyrych: I don't think this path fits to the controller. You should change it.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,37 @@
...
+@CrossOrigin(origins = "http://localhost:4200")
+@RequiredArgsConstructor
+@Slf4j
+public class ImageController {
+    private static final String NAME_OF_CLASS = "IN the ImageController: ";
```
rpkyrych: This constant is unused - just remove it.

src/main/java/com/hillel/items_exchange/dao/ImageRepository.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+
+@Repository
+public interface ImageRepository extends JpaRepository<Image, Long> {
+
+    @Query(value = "SELECT resource_url FROM image WHERE  product_id = ?1", nativeQuery = true)
+    List<String> findImageUrlsByProductId(Long id);
+
+    @Query(value = "SELECT id, product_id, resource_url, default_photo FROM image WHERE  product_id = ?1", nativeQuery = true)
```
rpkyrych: Queries into this repository are quite easy. You could get rid of them end use Spring Data Expression language to create requests with key words: 
https://docs.spring.io/spring-data/keyvalue/docs/current/reference/html/#repository-query-keywords

src/test/java/com/hillel/items_exchange/controller/ImageControllerTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class ImageControllerTest {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Test
+    @Transactional
+    @DataSet("database_init.yml")
+    void getImageUrlsByProductId_shouldReturnAllImageUrls() throws Exception {
+        mockMvc.perform(get("/products/1/imageUrls")
```
rpkyrych: You should use URL with method param and pass value directly into the test: ``mockMvc.perform(get("/products/{product_id}/imageUrls", 1L)``.
You should verify that expected quantity of the images is equals to values that you get into the tests.

src/main/java/com/hillel/items_exchange/dao/ImageRepository.java
lubchik-14: I wonder if we do need findImageUrlsByProductId(Long id) or not?

rpkyrych: Lest keep is for now.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
lubchik-14: I removed RequestMapping and changed GetMapping. I am not certain how to do it right.

src/main/java/com/hillel/items_exchange/dao/ImageRepository.java
lubchik-14: I rewrote those requests. As for me now they aren't easy enough to be removed.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,39 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+
+@RestController
```
rpkyrych: You should add @RequestMapping("/image") below the annotation.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,39 @@
...
+@RequiredArgsConstructor
+@Slf4j
+public class ImageController {
+    private final ImageService imageService;
+
+    @GetMapping("/adv/{advertisement_id}/product/{product_id}/image-urls")
```
rpkyrych: You get here images not advertisements so it should be '/image' global path here. I don't know why do you need advertisement here? Just use product and that's all and it would be like "/links/{product_id}".

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,39 @@
...
+    public ResponseEntity<List<String>> getImageUrlsByAdvertisementIdAndProductId(
+            @PathVariable("advertisement_id") long advId,
+            @PathVariable("product_id") long productId) {
+        List<String> imageUrls = imageService.getImageUrlsByAdvertisementIdAndProductId(advId, productId);
+        return (imageUrls.isEmpty())
+                ? new ResponseEntity<>(HttpStatus.NOT_FOUND)
+                : new ResponseEntity<>(imageUrls, HttpStatus.OK);
+    }
+
+    @GetMapping("/adv/{advertisement_id}/product/{product_id}/images")
```
rpkyrych: The same is here - get rid of non-useful advertisement just make it as "/{product_id}"

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,39 @@
...
+    private final ImageService imageService;
+
+    @GetMapping("/adv/{advertisement_id}/product/{product_id}/image-urls")
+    public ResponseEntity<List<String>> getImageUrlsByAdvertisementIdAndProductId(
+            @PathVariable("advertisement_id") long advId,
+            @PathVariable("product_id") long productId) {
+        List<String> imageUrls = imageService.getImageUrlsByAdvertisementIdAndProductId(advId, productId);
+        return (imageUrls.isEmpty())
+                ? new ResponseEntity<>(HttpStatus.NOT_FOUND)
+                : new ResponseEntity<>(imageUrls, HttpStatus.OK);
```
rpkyrych: Nice approach! :+1:

src/main/java/com/hillel/items_exchange/dao/ImageRepository.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.hillel.items_exchange.model.Image;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+
+@Repository
+public interface ImageRepository extends JpaRepository<Image, Long> {
```
rpkyrych: All requests might be simplified by using a product only.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,39 @@
...
+@RestController
+@RequiredArgsConstructor
+@Slf4j
+public class ImageController {
+    private final ImageService imageService;
+
+    @GetMapping("/adv/{advertisement_id}/product/{product_id}/image-urls")
+    public ResponseEntity<List<String>> getImageUrlsByAdvertisementIdAndProductId(
+            @PathVariable("advertisement_id") long advId,
+            @PathVariable("product_id") long productId) {
```
rpkyrych: You should add validation annotation for parameters or internal checking.

src/test/java/com/hillel/items_exchange/controller/ImageControllerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Test
+    @Transactional
+    void getByAdvertisementIdAndProductId_shouldReturnNotFound() throws Exception {
+        mockMvc.perform(get("/adv/{advertisement_id}/product/{product_id}/images", 1L, 2L)
+                .accept(MediaType.APPLICATION_JSON))
+                .andDo(print())
+                .andExpect(status().isNotFound());
+        mockMvc.perform(get("/adv/{advertisement_id}/product/{product_id}/images", 2L, 1L)
```
rpkyrych: One test condition here is enough.

src/test/java/com/hillel/items_exchange/controller/ImageControllerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Test
+    @Transactional
+    void getImageUrlsAdvertisementIdAndProductId_shouldReturnNotFound() throws Exception {
+        mockMvc.perform(get("/adv/{advertisement_id}/product/{product_id}/image-urls", 1L, 2L)
+                .accept(MediaType.APPLICATION_JSON))
+                .andDo(print())
+                .andExpect(status().isNotFound());
+        mockMvc.perform(get("/adv/{advertisement_id}/product/{product_id}/image-urls", 2L, 1L)
```
rpkyrych: One test condition it enough here.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -0,0 +1,47 @@
...
+    @GetMapping("/{product_id}")
+    public ResponseEntity<List<ImageDto>> getByProductId(@PathVariable("product_id")
+                                             @PositiveOrZero(message = "{invalid.id}") long id) {
+        List<ImageDto> images = imageService.getByProductId(id);
+        return (images.isEmpty())
+                ? new ResponseEntity<>(HttpStatus.NOT_FOUND)
+                : new ResponseEntity<>(images, HttpStatus.OK);
+    }
+
+    @ExceptionHandler(ConstraintViolationException.class)
```
rpkyrych, line 42: What is the case when current exception appears?

src/main/java/com/hillel/items_exchange/service/ImageService.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.stream.Collectors;
+
+@Service
+@RequiredArgsConstructor
+public class ImageService {
+    private final ModelMapper modelMapper;
+    private final ImageRepository imageRepository;
+
+    public List<String> getLinksByProductId(Long id) {
+        return imageRepository.findByProductId(id).stream()
```
rpkyrych, line 21: You can create new method into repository that will extract only links from image it would be equal to SQL request:
``select links from image;``

src/main/java/com/hillel/items_exchange/controller/ImageController.java
lubchik-14: It appears when product_id is less then 0.

PR #71, opened 2020-04-26
CGE-70 Create REST endpoint for getting user's data

21November:
Created REST endpoint /user/info/{id}
that return all user's data.

Only registered users could use this endpoint.
Users don't have possibility to view other users' information.
New code is covered by unit tests.


src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.security.Principal;
+
+@RestController
+@RequestMapping("/user")
+@CrossOrigin(origins = "http://localhost:4200")
+@RequiredArgsConstructor
+@Slf4j
```
rpkyrych, line 18: Why did you added the logger here and didn't use it?

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,37 @@
...
+public class UserController {
+
+    private final UserService userService;
+
+    @RequestMapping(value = "/info/{id}", method = RequestMethod.GET)
+    public @ResponseBody
+    ResponseEntity<UserDto> getUserDto(@PathVariable("id") Long id, Principal principal) {
```
rpkyrych, line 25: We use primitive types here to prevent null as id here...

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,37 @@
...
+
+        User user = userService.findByUsernameOrEmail(principal.getName()).get();
```
rpkyrych, line 27: You use Optional here so it's bad idea to extract user without check if it exist or you could get NullPointerException. So it could be easily changed something like that:
`        Optional<Long> userId = userService.findByUsernameOrEmail(principal.getName()).map(BaseEntity::getId);
        if (userId.isEmpty() || userId.get() != id){
            return new ResponseEntity<>(HttpStatus.FORBIDDEN);
        }`

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,37 @@
...
+
+        if (user != null && user.getId() != id){
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
rpkyrych, line 30: It's would be better to throw an exception with clarification that the user doesn't have access to review other user's data and FORBIDDEN code.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,37 @@
...
+@RestController
+@RequestMapping("/user")
+@CrossOrigin(origins = "http://localhost:4200")
+@RequiredArgsConstructor
+@Slf4j
+public class UserController {
+
+    private final UserService userService;
+
+    @RequestMapping(value = "/info/{id}", method = RequestMethod.GET)
```
rpkyrych, line 23: Could be simplified with GetMapping annotation.

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
```diff
@@ -93,4 +96,15 @@ public void registerNewUser(UserRegistrationDto userRegistrationDto, BCryptPassw
             log.warn("IN UserServiceImpl (registerNewUser): user: {} not registered", user);
         }
     }
+
+    @Override
+    public Optional<UserDto> getUserDtoById(Long id) {
+        return userRepository.findById(id).map(this::mapUserToDto);
```
rpkyrych, line 102: :+1:

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
```diff
@@ -0,0 +1,44 @@
...
+class UserControllerTest {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Test
+    @WithMockUser(username = "admin")
+    @Transactional
+    @DataSet("database_init.yml")
+    void getUserDto_shouldReturnUserDtoIfExists() throws Exception {
```
rpkyrych, line 35: Would you mind creation a negative test scenario as well?

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,45 @@
...
+    private final UserService userService;
+
+    @GetMapping("/info/{id}")
+    public @ResponseBody
+    ResponseEntity<UserDto> getUserInfo(@PathVariable("id") long id, Principal principal) {
+
+        userService.findByUsernameOrEmail(principal.getName())
+                .map(BaseEntity::getId)
+                .filter(userId -> userId == id)
+                .orElseThrow(() -> new AccessDeniedException("You do not have access to view the data of another user!"));
```
KozminAlexandr, line 31: Would you mind to extract message to property file?

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,45 @@
...
+
+        return userService.getUserDto(id)
+                .map(userDto -> new ResponseEntity<>(userDto, HttpStatus.OK))
+                .orElseGet(() -> new ResponseEntity<>(HttpStatus.NOT_FOUND));
```
KozminAlexandr, line 35: I think, that it would be more appropriate here to throw an exception with detailed information

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
```diff
@@ -0,0 +1,50 @@
...
+                .accept(MediaType.APPLICATION_JSON))
+                .andDo(print())
+                .andExpect(jsonPath("$.id").value(1))
+                .andExpect(status().isOk());
+
+        mockMvc.perform(get("/user/info/{id}", 2L)
+                .content(asJsonString("admin"))
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON))
+                .andDo(print());
```
KozminAlexandr, line 48: One test - one scenario. And i don't understand a bit, in the second scenario there is not enough verification of the results?

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
```diff
@@ -93,4 +96,15 @@ public void registerNewUser(UserRegistrationDto userRegistrationDto, BCryptPassw
             log.warn("IN UserServiceImpl (registerNewUser): user: {} not registered", user);
         }
     }
+
+    @Override
+    public Optional<UserDto> getUserDto(Long id) {
+        return userRepository.findById(id).map(this::mapUserToDto);
```
lubchik-14, line 102: `lastOnlineTime` becomes `Null` after that mapping. I think something is wrong.

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: Done. Check it out, please.

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: Can I give you more details? How do you imagine that?

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
21November: I guess you're right. Didn't get it figured out.

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
21November: Honestly, I don't know the right answer. The test does show that the field lastOnlineTime outputs with the value null

21November: Do you know how to solve such a problem?

lubchik-14: I think the problem is about types of lastOnlineTine in UserDto and User. They are different, so mapper does something wrong mapping that field

21November: That's right, they're different. Then all you have to do is bring them to the same type. I wonder what Roma has to say about that. Thank you for your comment))

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,48 @@
...
+@RequiredArgsConstructor
+@Slf4j
+public class UserController {
+
+    private final MessageSource messageSource;
+    private final UserService userService;
+
+    @GetMapping("/info/{id}")
+    public @ResponseBody
+    ResponseEntity<UserDto> getUserInfo(@PathVariable("id") long id, Principal principal) {
```
lubchik-14, line 29: In my opinion `UserDto` doesn't cover all user's data

21November: I agree with you. But that's the task. If the task is changed to return all information to the user, I will think about how to return absolutely all the information.

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
21November: For fun, I changed the LocalDate type to LocalDateTime in UserDto and now the lastOnlineTime field returns correctly

lubchik-14: If I were you, I would discuss this issue with Roma and fix it in your PR.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,48 @@
...
+
+        userService.findByUsernameOrEmail(principal.getName())
+                .map(BaseEntity::getId)
+                .filter(userId -> userId == id)
+                .orElseThrow(() -> new AccessDeniedException(messageSource.getMessage("invalid.token.is.not.equal.to.your.token", null, Locale.getDefault())));
```
rpkyrych, line 34: You definitely should make the message variable name quite shorter.

src/main/java/com/hillel/items_exchange/controller/UserController.java
rpkyrych: @21November would you mind adding validation annotation to id variable?

src/main/java/com/hillel/items_exchange/controller/UserController.java
rpkyrych: @21November I propose you to change what you've done here in such way:
1. To create additional method into the service layer ``Optional<UserDto> getByUsernameOrEmail`` that will return you DTO. Under the hood you will use userRepository.findByUsername(usernameOrEmail) to get those information.
2. After getting your UserDto you should compare it's id with that that you received into ``getUserInfo(@PathVariable("id") long id``.
3. Return that user if id equals and return AccessDenied exception with description or 403 (Forbidden) code.
Doing that way you decrease quantity of requests to the database and make the code simpler.

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
rpkyrych: @21November do nothing with `last online` for now I'll create ticket to populate this field.

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: Added some changes. Take a look, please.
The method now addresses the database once.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private final MessageSource messageSource;
+    private final UserService userService;
+
+    @GetMapping("/info/{id}")
+    public @ResponseBody
+    ResponseEntity<UserDto> getUserInfo(@Validated @PathVariable("id") Long id, Principal principal) {
+
+        UserDto userDto = userService.getByUsernameOrEmail(principal.getName()).orElse(new UserDto());
+        if (!id.equals(userDto.getId())){
```
lubchik-14, line 31: I can't understand why you have created a new instance of UserDto and then compare ids

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
```diff
@@ -93,4 +96,15 @@ public void registerNewUser(UserRegistrationDto userRegistrationDto, BCryptPassw
             log.warn("IN UserServiceImpl (registerNewUser): user: {} not registered", user);
         }
     }
+
+    @Override
+    public Optional<UserDto> getByUsernameOrEmail(String usernameOrEmail) {
+        return userRepository.findByUsername(usernameOrEmail).map(this::mapUserToDto);
+    }
```
lubchik-14, line 103: I'm sorry that didn't mentioned earlier, but I think this method should returns `Optional<User>`. I can't use the return type UserDto, it doesn't fit to my logic.

src/main/resources/messages.properties
```diff
@@ -10,6 +10,7 @@ invalid.password=Your password must be between 8 and 30 characters at must conta
 invalid.username.size=Your login must be between 2 and 50 characters, do not use space
 invalid.token=You must be authenticated
 token.not.start.with.bearer=JWT token is not starts with Bearer prefix
+token.not.equal.to.your.token=You do not have access to view the data of another user!
```
lubchik-14, line 13: You don't use that message anywhere

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,45 @@
...
+@RequiredArgsConstructor
+@Slf4j
+public class UserController {
+
+    private final MessageSource messageSource;
+    private final UserService userService;
+
+    @GetMapping("/info/{id}")
+    public @ResponseBody
+    ResponseEntity<UserDto> getUserInfo(@Validated @PathVariable("id") Long id, Principal principal) {
```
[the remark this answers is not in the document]
21November: Used Long to apply Id check by equals (!id.equals(userDto.getId()))

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: I did this to reduce the number of hits to the database. This way I already know the information about the user and if his id is equal to the input Id, then the user is given the information

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
21November: I heard you. I'll talk to Roma about it.

src/main/resources/messages.properties
21November: I know it doesn't apply. I'm a little confused as to where to apply it now or where to remove it at all. Since I was given a choice above to either output an informative message or ERROR 403.

src/main/java/com/hillel/items_exchange/controller/UserController.java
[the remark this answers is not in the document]
lubchik-14: id != userDto.getId() still works or not 🤔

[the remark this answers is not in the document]
21November: Yes, it works. But I thought it was more reliable. It won't take long to get back to the mark.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,45 @@
...
+import java.security.Principal;
+
+@RestController
+@RequestMapping("/user")
+@CrossOrigin(origins = "http://localhost:4200")
+@RequiredArgsConstructor
+@Slf4j
+public class UserController {
+
+    private final MessageSource messageSource;
```
rpkyrych, line 23: Remove it - you don't use it here.

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
```diff
@@ -93,4 +96,15 @@ public void registerNewUser(UserRegistrationDto userRegistrationDto, BCryptPassw
             log.warn("IN UserServiceImpl (registerNewUser): user: {} not registered", user);
         }
     }
+
+    @Override
+    public Optional<UserDto> getByUsernameOrEmail(String usernameOrEmail) {
+        return userRepository.findByUsername(usernameOrEmail).map(this::mapUserToDto);
```
rpkyrych, line 102: :+1:

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
rpkyrych: @lubchik-14 you can do with Optional whatever you want. :smiley:

src/main/resources/messages.properties
rpkyrych: You might use it as an exception message using MessageSource.

src/main/java/com/hillel/items_exchange/controller/UserController.java
lubchik-14: Maybe I'm mistaken, but you hit to the DB once taking User by name or nothing, so then you can work with that instance without any hitting to DB

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
lubchik-14: @rpkyrych 🤔 but UserDto does not have for example children or phones. And I have to work with User instance and its relations. How do I do this? Can you give me an example, please.

src/main/java/com/hillel/items_exchange/controller/UserController.java
rpkyrych: @21November the question remains. :arrow_up:

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,46 @@
...
+@RequestMapping("/user")
+@RequiredArgsConstructor
+@Slf4j
+public class UserController {
+
+    private final MessageSource messageSource;
+    private final UserService userService;
+
+    @GetMapping("/info/{id}")
+    public @ResponseBody
```
rpkyrych, line 27: You could get rid of ResponseBody annotation. I had written about it to the chat few days ago.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,46 @@
...
+    ResponseEntity<UserDto> getUserInfo(@Validated @PathVariable("id") long id, Principal principal) {
```
rpkyrych, line 28: Why do you use Validated annotation here?

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -0,0 +1,46 @@
...
+        return ResponseEntity.ok(userService.getByUsernameOrEmail(principal.getName())
+                .filter(user -> user.getId() == id)
+                .orElseThrow(() -> new AccessDeniedException(messageSource.getMessage(
+                        "token.not.equal.to.your.token",
+                        null,
+                        Locale.getDefault()))
+                ));
+    }
+
+    @ExceptionHandler(AccessDeniedException.class)
```
rpkyrych, line 39: You should merge master branch to yours and move this part to GlobalExceptionHandler.

src/main/java/com/hillel/items_exchange/service/impl/UserServiceImpl.java
rpkyrych: Discussed internally with @lubchik-14. The topic is resolved.

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: At one point, the logger was used. Then the code was changed and the logger was no longer needed. I'm gonna delete this annotation.

21November: line 41 of this code uses logger. 

log.warn(e.getMessage(), e);

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: annotation "ResponseBody" has been deleted

src/main/java/com/hillel/items_exchange/controller/UserController.java
21November: Earlier, you asked to perform id validation with Spring. That's what I did. My choice isn't right?

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
```diff
@@ -121,6 +122,14 @@
         return new ResponseEntity<>(errorMessage, HttpStatus.BAD_REQUEST);
     }
 
+    @ExceptionHandler(AccessDeniedException.class)
+    public ResponseEntity<String> handlerAccessDenied(AccessDeniedException e) {
```
KozminAlexandr, line 126: Would you mind to refactor this handler to make it look like the rest? You can use ErrorMassage, extract `.body(...)` to property file and log error with `logErrorMessage()`

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
```diff
@@ -0,0 +1,62 @@
...
+@DBRider
+@AutoConfigureMockMvc
+@Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, scripts = "classpath:index-reset.sql")
+class UserControllerTest {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Test
+    @WithMockUser(username = "admin")
```
KozminAlexandr, line 37: Can you explain, please, why need `username = "admin"`?

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
rpkyrych: @KozminAlexandr I've just moved the code to current place instead from the controller. Current ticket blocks 2 other tickets that @lubchik-14 and @drw85 are working on.
If you think that it's on time and you can do it very quickly - go ahead!

KozminAlexandr: @rpkyrych If you do not mind, I will do it in my next ticket

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
rpkyrych: We mock user here to get it from Principal into the controller.

PR #74, opened 2020-05-31
CGE-85: flow with images

rpkyrych:
Changed image links to real image files.


src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -1,43 +1,99 @@
...
     private final ImageService imageService;
+    private final UserService userService;
 
-    @GetMapping("/{product_id}/links")
-    public ResponseEntity<List<String>> getImageLinksByProductId(@PathVariable("product_id")
-                                                                @PositiveOrZero(message = "{invalid.id}") long id) {
-        List<String> links = imageService.getLinksByProductId(id);
-        return (links.isEmpty())
+    @GetMapping(value = "/{product_id}/resource")
+    public ResponseEntity<List<byte[]>> getImageLinksByProductId(@PathVariable("product_id")
```
KozminAlexandr, line 39: In my humble opinion, it would be nice to rename this controller

rpkyrych: I think you're saying about the method's name. Yup, I missed it and rename it. Thanks.

KozminAlexandr: Yes, about method =)

src/main/java/com/hillel/items_exchange/model/Image.java
```diff
@@ -15,8 +16,9 @@
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private long id;
-    @Column(name = "resource_url")
-    private String resourceUrl;
+    @Lob
+    @Type(type="org.hibernate.type.BinaryType")
+    private byte[] resource;
```
KozminAlexandr, line 21: Annotation @Column is not needed for this field?

src/main/java/com/hillel/items_exchange/service/ImageService.java
```diff
@@ -3,23 +3,28 @@
...
 
 @Service
 @RequiredArgsConstructor
 public class ImageService {
     private final ModelMapper modelMapper;
+    private final ProductService productService;
     private final ImageRepository imageRepository;
 
-    public List<String> getLinksByProductId(Long id) {
+    public List<byte[]> getImageResources(Long id) {
```
KozminAlexandr, line 25: It would be nice to add to method name "ByProductId", because a little confusing

src/main/java/com/hillel/items_exchange/service/ProductService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.hillel.items_exchange.dao.ProductRepository;
+import com.hillel.items_exchange.model.Product;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+import java.util.Optional;
+
+@RequiredArgsConstructor
+@Component
```
KozminAlexandr, line 11: I understand, that @Service = @Component, but would you like to change this for better comprehensibility? =)

src/main/java/com/hillel/items_exchange/service/ProductService.java
```diff
@@ -0,0 +1,18 @@
...
+public class ProductService {
+    private final ProductRepository productRepository;
+
+    Optional<Product> findById(Long id) {
```
KozminAlexandr, line 15: What is the best practice: "unpacking" Optional on service layer or later in controller?

src/main/java/com/hillel/items_exchange/model/Image.java
rpkyrych: It is not needed when the name of the column into the database is equal to the variable's name.

src/main/java/com/hillel/items_exchange/service/ProductService.java
rpkyrych: Yup, I agree and I'll change it.

src/main/java/com/hillel/items_exchange/service/ProductService.java
rpkyrych: It depends of your purposes. Mostly Optional returns to the place where it is needed, so I prefer to send it to Controllers layer.

PR #75, opened 2020-06-13
Cge 88 update personal information

lubchik-14:
Add possibility to user add/update personal information at Personal Area


src/main/java/com/hillel/items_exchange/dao/UserRepository.java
```diff
@@ -1,13 +1,14 @@
...
 import com.hillel.items_exchange.model.User;
+import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.stereotype.Repository;
 
 import java.util.Optional;
 
 @Repository
-public interface UserRepository extends CrudRepository<User, Long> {
+public interface UserRepository extends CrudRepository<User, Long>, JpaRepository<User, Long> {
```
KozminAlexandr, line 11: Interface `JpaRepository` extends `PagingAndSortingRepository` which in turn extends `CrudRepository`. So I think, you can leave only `JpaRepository<User, Long>`

lubchik-14: OK, thanks!

src/main/java/com/hillel/items_exchange/dto/PhoneDto.java
```diff
@@ -0,0 +1,26 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+public class PhoneDto {
+    @PositiveOrZero(message = "{invalid.id}")
+    private long id;
+    @NotNull(message = "{invalid.not-null}")
+    private Boolean show;
+    @Pattern(regexp = "^\\s*(?<country>\\+?\\d{1,3})[-. (]*(?<area>\\d{3})[-. )]*(?<number>\\d{3}[-. ]*\\d{2}[-. ]*\\d{2})\\s*$",
```
KozminAlexandr, line 21: What you think about extraction regular expressions to property file? It seems to me, this will give quick access to them in case of changes

src/main/java/com/hillel/items_exchange/dto/UserDto.java
```diff
@@ -27,17 +29,17 @@
...
     @Size(min = 2, max = 50, message = "{invalid.size}")
     private String lastName;
-    @NotNull(message = "{invalid.not-null}")
     private String avatarImage;
     @NotNull(message = "{invalid.not-null}")
     @PastOrPresent(message = "{invalid.past-or-present.date}")
     @JsonFormat(pattern = "yyyy-MM-dd")
     private LocalDate lastOnlineTime;
-
+    @NotNull(message = "{invalid.not-null}")
```
KozminAlexandr, line 41: Hm.. Why user's lastName or avatarImage can be null, but phone or children can't?

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
```diff
@@ -123,11 +124,21 @@
     }
 
     @ExceptionHandler(AccessDeniedException.class)
-    public ResponseEntity<String> handlerAccessDenied(AccessDeniedException e) {
-        log.warn(e.getMessage(), e);
-        return ResponseEntity
-                .status(HttpStatus.FORBIDDEN)
-                .body("Exception during request!\nError message:\n" + e.getLocalizedMessage());
+    public ResponseEntity<ErrorMessage> handlerAccessDenied(AccessDeniedException e,
```
KozminAlexandr, line 127: 👍

src/main/java/com/hillel/items_exchange/model/User.java
```diff
@@ -48,9 +48,15 @@
             inverseJoinColumns = {@JoinColumn(name = "deal_id", referencedColumnName = "id")})
     private List<Deal> deals;
 
-    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL)
-    private List<Phone> phones;
+    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL, orphanRemoval = true)
+    private List<Phone> phones = new ArrayList<>();
```
KozminAlexandr, line 52: Why needs to initialize list here?

src/main/java/com/hillel/items_exchange/dto/PhoneDto.java
lubchik-14: I tried to find out about it, but was unsuccessful. And our project does not have any examples. So I did it according to existing code. Ok, I agree with you, and I will certain try to figure out it again

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
```diff
@@ -59,4 +94,162 @@ void negativeTestReceivingInformationAboutAnotherUser() throws Exception {
...
+    }
+
+    @Test
+    @Transactional
+    @DataSet("database_init.yml")
+    @ExpectedDataSet(value = "user/update_phone.yml", ignoreCols = {"last_online_time", "updated"})
+    void updateUserInfo_shouldUpdateUserPhones() throws Exception {
+        UserDto validUserDto = UserDtoUpdatingUtil.getValidUserDto();
+        validUserDto.getPhones().get(0).setPhoneNumber("+1 (234) 111-11-11");
+        validUserDto.getPhones().add(new PhoneDto(0, Boolean.FALSE, "+38.050.222.22.22", Boolean.FALSE));
```
KozminAlexandr, line 161: Maybe this can be simplified? Using list with not valid phones for example or something else, in order not to duplicate the code

src/main/java/com/hillel/items_exchange/dto/UserDto.java
lubchik-14: I followed db constraints when added those validations. In case with children or phone, we have an empty array. But I will try to deal with null and may remove those validation from children and phones.

KozminAlexandr: Need to check this moment with @rpkyrych. It's just business logic, but I think this entities may not be nulls.

src/main/java/com/hillel/items_exchange/dao/UserRepository.java
KozminAlexandr: If you delete `CrudRepository` it will not ruin this code =)

src/main/java/com/hillel/items_exchange/model/User.java
lubchik-14: Actually I can't explain why :). But according to Hibernate user guide examples I added it.
https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#associations-one-to-many

src/test/java/com/hillel/items_exchange/controller/UserControllerTest.java
lubchik-14: Firstly those phones are valid. And I did to show that masks of regexp are working good. But if somebody is agree with you, I will add them to List.

src/main/java/com/hillel/items_exchange/model/User.java
```diff
@@ -59,4 +59,14 @@ public void addUser() {
...
+
+    public void addChild(Child child) {
+        children.add(child);
+        child.setUser(this);
+    }
+
+    public void addPhone(Phone phone) {
+        phones.add(phone);
+        phone.setUser(this);
+    }
```
Jack11M, line 71: In my humble opinion. it's not a good idea to use methods (addPhone, addChild) here

src/main/java/com/hillel/items_exchange/model/User.java
```diff
@@ -54,17 +54,7 @@
...
-    public void addPhone(Phone phone) {
-        phones.add(phone);
-        phone.setUser(this);
-    }
-
-    @PrePersist
+    @PreUpdate
     public void addUser() {
         phones.forEach(phone -> phone.setUser(this));
         children.forEach(child -> child.setUser(this));
```
Jack11M, line 60: Your code works without this method, doesn't it?

src/main/java/com/hillel/items_exchange/service/ChildService.java
```diff
@@ -0,0 +1,37 @@
...
+                        .filter(c -> c.getId() == child.getId())
+                        .findAny()
+                        .orElseThrow(() -> new UnprocessableEntityException(
+                                messageSourceUtil.getExceptionMessageSource("exception.invalid.dto")));
+                BeanUtils.copyProperties(child, existedChild, "id", "user");
+            }
+        });
+        currentUser.getChildren().removeIf(child ->
+                updatedChildren.stream().map(Child::getId).noneMatch(id -> id == child.getId()));
+    }
```
Jack11M, line 35: I think you can simplify this method.

src/main/java/com/hillel/items_exchange/service/PhoneService.java
```diff
@@ -0,0 +1,37 @@
...
+                        .filter(p -> p.getId() == phone.getId())
+                        .findAny()
+                        .orElseThrow(() -> new UnprocessableEntityException(
+                                messageSourceUtil.getExceptionMessageSource("exception.invalid.dto")));
+                BeanUtils.copyProperties(phone, existedPhone, "id", "user");
+            }
+        });
+        currentUser.getPhones().removeIf(phone ->
+                updatedPhones.stream().map(Phone::getId).noneMatch(id -> id == phone.getId()));
+    }
```
Jack11M, line 35: If you add PhoneRepository, you can do something like this (just my opinion):

public void updateAll(List<Phone> newPhoneList, User user) {
    newPhoneList.forEach(phone -> {
        if (phone.getId() == 0) {
            phoneRepository.saveAndFlush(phone);
            user.addPhone(phone);
        } else {
            Phone existedPhone = phoneRepository.findById(phone.getId()).orElseThrow(() ->
                    new UnprocessableEntityException(messageSourceUtil.getExceptionMessageSource("exception.invalid.dto")));
            BeanUtils.copyProperties(phone, existedPhone, "id", "user");
            phoneRepository.saveAndFlush(existedPhone);
        }
    });
    user.getPhones().removeIf(phone -> !newPhoneList.contains(phone));

src/main/java/com/hillel/items_exchange/model/User.java
lubchik-14: Yes, you are right. I have forgotten about it.

src/main/java/com/hillel/items_exchange/model/User.java
lubchik-14: I fallowed the Hibernate user guid to synchronize both ends a bidirectional association.
https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#associations-one-to-many-bidirectional

src/main/java/com/hillel/items_exchange/service/ChildService.java
lubchik-14: Ok, I will try, if this method lives in the code.

src/main/java/com/hillel/items_exchange/service/PhoneService.java
lubchik-14: Yes, your example works, but imagine we got there UnprocessableEntityException here and operation of updating is interrupted. The previous code made changes (if they were) in User entity, Child table, so we have no transaction while updating user info.
I tried to save all entities saving User entity, but not erlier.

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -36,6 +43,30 @@ public boolean registerNewUser(UserRegistrationDto userRegistrationDto, Role rol
         return userRepository.save(registeredUser).getId() != 0;
     }
 
+    public UserDto update(UserDto newUserDto) {
+        User currentUser = userRepository.findById(newUserDto.getId())
+                .orElseThrow(IllegalStateException::new);
+        User updatedUser = mapDtoToUser(newUserDto);
+
+        if (!bCryptPasswordEncoder.matches(updatedUser.getPassword(), currentUser.getPassword())) {
+            throw new IllegalOperationException(
```
rpkyrych, line 52: If you throw something inside the service layer it must be checked exception.

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -51,4 +82,16 @@ public boolean existsByEmail(String email) {
     private UserDto mapUserToDto(User user) {
         return modelMapper.map(user, UserDto.class);
     }
+
+    private User mapDtoToUser(UserDto userDto) {
+        Converter<String, Long> stringLongConverter = new AbstractConverter<>() {
+            protected Long convert(String string) {
+                return Long.parseLong(string.replaceAll("[^\\d]", ""), 10);
```
rpkyrych, line 89: Really nice approach. :+1:

src/main/resources/application.properties
```diff
@@ -19,6 +19,7 @@ spring.liquibase.change-log=classpath:liquibase/db.changelog-master.xml
 ### Hibernate Properties
 # The SQL dialect makes Hibernate generate better SQL for the chosen database
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL8Dialect
+spring.jpa.properties.hibernate.globally_quoted_identifiers=true
```
rpkyrych, line 22: In which cases you get a problem that this property covers?

src/main/resources/messages.properties
```diff
@@ -52,5 +53,9 @@ exception.invalid.dto=Invalid data transfer object
 exception.sql=Exception during saving object to the database
 exception.entity.not-found=Entity not found into the database
 exception.illegal.argument=Exception during request. Illegal or inappropriate argument
+exception.illegal.operation=Operation is not allowed
+exception.illegal.password.change=You are unable to change your password
+exception.illegal.username.change=You are unable to change your username
```
rpkyrych, line 58: You should make it universal:
`exception.illegal.field.change=You are unable to change your ${validatedValue}`

Jack11M:
> Thanks for review!
> I don't understand what the reason is in your advice of removing "orphanRemoval = true" for Children and Phones.
> Also, unfortunately, I can't understand about tests. Would you explain to me which code does not cover which test cases?
> And yes, I agree my putUserDto_shouldBeThrownUnprocessableEntityException() should rename like putUserDto_shouldBeThrownUnprocessableEntityExceptionIfChildHasWrongUserId(). So now we have 100% of covering all services.

Jack11M:
> > Thanks for review!
> > I don't understand what the reason is in your advice of removing "orphanRemoval = true" for Children and Phones.
> > Also, unfortunately, I can't understand about tests. Would you explain to me which code does not cover which test cases?
> > And yes, I agree my putUserDto_shouldBeThrownUnprocessableEntityException() should rename like putUserDto_shouldBeThrownUnprocessableEntityExceptionIfChildHasWrongUserId(). So now we have 100% of covering all services.

Jack11M:
> In theory, you can avoid some method in User class and simplify your code a bit.
> For example try checking the following code in UserController. But firstly you should remove "orphanRemoval = true" for Children and Phones.
> Also this code does not cover some test case (putUserDto_shouldBeThrownUnprocessableEntityException()
> putUserDto_shouldBeThrownUnprocessableEntityExceptionIfPhoneHasWrongUserId()).
> 
> BeanUtils.copyProperties(updatedUser, currentUser,
> "id", "advertisements", "deals", "role", "password", "online", "lastOnlineTime",
> "created", "updated", "status");
> currentUser.getPhones().forEach(phone -> phone.setUser(currentUser));
> currentUser.getChildren().forEach(child -> child.setUser(currentUser));

I mean, you can try to optimize my code in this comment to avoid using Children and Phone Services, but then you need to remove “orphanRemoval = true” for Children and Phones. But this is not relevant now, because @rpkyrych offers a better approach.

src/main/java/com/hillel/items_exchange/dto/UserDto.java
lubchik-14: All of those NotNull constraints make a user to add/change all his information. Right after the user registration avatarImage, firstName, lastName are empty string. So if he wants add first name only (actually change value of firstName from "" to user first name), he won't be able to do it because of lastName and avatarImage can't be "".
According to Declaration of thsi task I have to give a possibility to add only what a user wants and edit it if he wants. 
So I am going to create a custom annotation to have "" is valid, null isn't. Am I on the rigth way?

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -43,4 +42,16 @@
                         Locale.getDefault()))
                 ));
     }
+
+    @PutMapping("/info/{id}")
+    public ResponseEntity<UserDto> updateUserInfo(@Valid @RequestBody UserDto userDto,
+                                                  @PositiveOrZero @PathVariable("id") long id,
```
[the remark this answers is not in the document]
lubchik-14: Actually, if you think it is OK with REST, I will get rid of this path variable.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -43,4 +42,16 @@
...
+                                                  @AuthenticationPrincipal JwtUser user) {
+        Long userId = user.getId();
+        if (userId != id || userId != userDto.getId()) {
+            throw new AccessDeniedException(
+                    messageSourceUtil.getExceptionMessageSource("exception.permission-denied.user-profile"));
```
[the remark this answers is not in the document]
lubchik-14: Ok

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
```diff
@@ -123,11 +124,21 @@
...
+                                                            ServletWebRequest request) {
+        ErrorMessage errorMessage = getErrorMessage(request, HttpStatus.FORBIDDEN,
+                "exception.security", Collections.singletonList(e.getLocalizedMessage()));
```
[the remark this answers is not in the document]
lubchik-14: Did you mean to add my exception **message**? Ok, I'll try.

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -36,6 +43,30 @@ public boolean registerNewUser(UserRegistrationDto userRegistrationDto, Role rol
...
+        User currentUser = userRepository.findById(newUserDto.getId())
+                .orElseThrow(IllegalStateException::new);
+        User updatedUser = mapDtoToUser(newUserDto);
+
+        if (!bCryptPasswordEncoder.matches(updatedUser.getPassword(), currentUser.getPassword())) {
+            throw new IllegalOperationException(
+                    messageSourceUtil.getExceptionMessageSource("exception.illegal.password.change"));
+        }
+
+        if (!updatedUser.getUsername().equals(currentUser.getUsername())) {
```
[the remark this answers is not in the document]
lubchik-14: Previously you asked me to throw an exception if password is changed. And I did it. Unfortunately I have no experience in conversations with programmers, so I don't undestand what I should do exactly.
Shold I ignore password at all, or remove password field from UserDto?

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -36,6 +43,30 @@ public boolean registerNewUser(UserRegistrationDto userRegistrationDto, Role rol
         return userRepository.save(registeredUser).getId() != 0;
     }
 
+    public UserDto update(UserDto newUserDto) {
+        User currentUser = userRepository.findById(newUserDto.getId())
+                .orElseThrow(IllegalStateException::new);
+        User updatedUser = mapDtoToUser(newUserDto);
```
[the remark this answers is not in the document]
lubchik-14: If I exclude this field, two object below will be equal
Phone(1, 380501111111, false, false, null)
Phone(0, 380502222222, false, false, null)
So after mapping and collection Set of phones contains only first one.
What is the reason of excluding "phoneNumber"?

src/main/resources/application.properties
lubchik-14: In case when SQL query contains a keyword "show". It can be solved by adding @Column(name = "\"show\"") to field of Phone class, but then tests are failed. Maybe because of h2 in tests, but for testing this field should not be quote, so I didn't add this property to test/resources/application.property. 
Also I think not using keywords can help.

src/main/java/com/hillel/items_exchange/service/UserService.java
lubchik-14: Thanks

src/main/java/com/hillel/items_exchange/dto/UserDto.java
rpkyrych: Discussed internally - @lubchik-14 will additionally check the problem.

src/main/java/com/hillel/items_exchange/controller/UserController.java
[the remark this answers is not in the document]
rpkyrych: Let's do that - just remove id from request.

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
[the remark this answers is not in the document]
lubchik-14: I'm sorry. I misunderstood you first. But existed hendler returns 409 status and my does 403. Is it ok to return 409 status in both exceptions? Or should I check it in that handler and return or 409, or 403?

src/main/java/com/hillel/items_exchange/util/PatternHandler.java
```diff
@@ -0,0 +1,15 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class PatternHandler {
+
+    public static final String EMAIL = "^[\\w-\\+]+(\\.[\\w]+)*@[\\w-]+(\\.[\\w]+)*(\\.[a-zA-Z]{2,})$";
+    public static final String PASSWORD_MIN_7_MAX_30 = "(?=.*?[0-9])(?=.*?[a-z])(?=.*?[A-Z])(?=\\S+$).+.{7,30}";
+    public static final String PHONE_NUMBER =
+            "^\\s*(?<country>\\+?\\d{2})[-. (]*(?<area>\\d{3})[-. )]*(?<number>\\d{3}[-. ]*\\d{2}[-. ]*\\d{2})\\s*$";
+    public static final String USERNAME_MIN_2_MAX_50 = "(?=\\S+$).{2,50}";
+    public static final String WORD_EMPTY_OR_MIN_2_MAX_50 = "^$|^[\\w-']{2,50}+$";
+}
```
Jack11M, line 15: Great approach!

src/main/java/com/hillel/items_exchange/model/User.java
```diff
@@ -48,9 +48,15 @@
             inverseJoinColumns = {@JoinColumn(name = "deal_id", referencedColumnName = "id")})
     private List<Deal> deals;
 
-    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL)
-    private List<Phone> phones;
+    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL, orphanRemoval = true)
+    private Set<Phone> phones;
```
Jack11M, line 52: Perhaps it is better to use the collection here, in case we will need a List in the future?

lubchik-14: Actually I had to undo this changes, because I don't use this collection any more in this task. I think it is better not to make changes in User class at all. What do you think about it?

src/main/java/com/hillel/items_exchange/model/User.java
```diff
@@ -48,9 +48,15 @@
...
 
-    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL)
-    private List<Child> children;
+    @OneToMany(mappedBy = "user", cascade = CascadeType.ALL, orphanRemoval = true)
+    private Set<Child> children;
+
+    @PreUpdate
+    public void addUser() {
+        phones.stream().filter(phone -> phone.getUser() == null).forEach(phone -> phone.setUser(this));
```
rpkyrych, line 59: Todo: remove this method.

src/main/resources/liquibase/db.changelog-1.0.xml
```diff
@@ -440,4 +440,10 @@
                     type="varchar(15)"/>
         </addColumn>
     </changeSet>
+
+    <changeSet id="1.0.26: NotNull constraint set in table phone" author="lubchik">
+        <addNotNullConstraint tableName="phone"
```
rpkyrych, line 445: Great! :tada:

src/main/java/com/hillel/items_exchange/mapper/UtilMapper.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.stream.Collectors;
+
+@Component
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class UtilMapper {
+
+    private static ModelMapper mapper;
+
+    @Autowired
+    public void setModelMapper(ModelMapper modelMapper) {
```
rpkyrych, line 24: Make the method static too.

PR #76, opened 2020-06-18
CGE-90: Fix update advertisement

KozminAlexandr:
Fixed advertisement update (child entities updated too)

### DIFF
--- a/src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
+++ b/src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
@@ -1,0 +1,13 @@
+package com.hillel.items_exchange.dao;
+
+import com.hillel.items_exchange.model.Location;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+@Repository
+public interface LocationRepository extends JpaRepository<Location, Long> {
+
+    Optional<Location> findLocationByCityAndDistrict(String city, String district);
+}
--- a/src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
+++ b/src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
@@ -9,5 +9,5 @@ import java.util.Optional;
 @Repository
 public interface LocationRepository extends JpaRepository<Location, Long> {
 
-    Optional<Location> findLocationByCityAndDistrict(String city, String district);
+    Optional<Location> findByCityAndDistrict(String city, String district);
 }
--- a/src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
+++ b/src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
@@ -4,10 +4,7 @@ import com.hillel.items_exchange.model.Location;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
-import java.util.Optional;
-
 @Repository
 public interface LocationRepository extends JpaRepository<Location, Long> {
-
-    Optional<Location> findByCityAndDistrict(String city, String district);
+    
 }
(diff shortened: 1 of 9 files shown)


src/main/java/com/hillel/items_exchange/service/AdvertisementService.java
```diff
@@ -54,6 +55,11 @@ public AdvertisementDto createAdvertisement(AdvertisementDto advertisementDto, U
         adv.setStatus(Status.NEW);
         adv.getProduct().setSubcategory(subcategoryService.findById(advertisementDto.getProduct().getSubcategoryId())
                 .orElseThrow(EntityNotFoundException::new));
+
+        Optional<Location> location = locationService.findByCityAndDistrict(
+                adv.getLocation().getCity(), adv.getLocation().getDistrict());
+        adv.setLocation(location.orElseGet(() -> locationService.createLocation(adv.getLocation())));
+
```
Jack11M: I remember that you have discussed the logic of adding Location if it does not exist, but what about the case when we, for example, receive the city "Kharkiv" in the first advertisement and city "Kharkov" in the second (for now we create new Location). I think we need to discuss adding a new Location one more time in the next meeting.

src/main/java/com/hillel/items_exchange/dao/LocationRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import com.hillel.items_exchange.model.Location;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+@Repository
+public interface LocationRepository extends JpaRepository<Location, Long> {
+
+    Optional<Location> findLocationByCityAndDistrict(String city, String district);
```
Jack11M, line 12: As for me, it's better to rename this method to "findByCityAndDistrict(String city, String district)" because according to the "Spring Data JPA documentation" in the methods of JpaRepository not duplicate the name of returns entity (findById, findByName, etc).

src/main/java/com/hillel/items_exchange/service/AdvertisementService.java
```diff
@@ -52,24 +51,64 @@ public AdvertisementDto createAdvertisement(AdvertisementDto advertisementDto, U
...
+                .orElseThrow(EntityNotFoundException::new));
+    }
+
+    private void updateImages(Product toUpdateProduct, Product fromDBProduct) {
+        List<Image> existImages = fromDBProduct.getImages();
+        List<Image> newImages = toUpdateProduct.getImages();
+        if (existImages.size() != newImages.size() || !existImages.containsAll(newImages)) {
+            existImages.retainAll(newImages);
+        }
+    }
```
lubchik-14, line 101: In fact this `updateImages` method doesn't update images, it only removes elements from the `existImages` that are not contained in the `newImages`. Changing field value defaultPhoto will be prosseced within another task?

src/main/java/com/hillel/items_exchange/service/AdvertisementService.java
```diff
@@ -52,24 +51,64 @@ public AdvertisementDto createAdvertisement(AdvertisementDto advertisementDto, U
...
-                .orElseThrow(EntityNotFoundException::new));
+        updateSubcategory(adv.getProduct(), advertisementDto.getProduct().getSubcategoryId());
+
         return mapAdvertisementToDto(advertisementRepository.save(adv));
     }
 
     public AdvertisementDto updateAdvertisement(AdvertisementDto dto) {
         Advertisement toUpdate = mapDtoToAdvertisement(dto);
         Advertisement fromDB = advertisementRepository.findById(dto.getId())
                 .orElseThrow(EntityNotFoundException::new);
```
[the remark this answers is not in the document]
rpkyrych: Yup, you're absolutely right. As you might see I don't change current logic and all those changes were created by me very long time ago mostly as temporal solution when we were expected lots of hibernate configuration issues.
I don't want to change such part of the code into current pull-request. It's better to create another ticket for removing such exception from all parts of the applicaton.

src/main/java/com/hillel/items_exchange/service/AdvertisementService.java
rpkyrych: Yup, I missed that part, thanks. I'll update it.

src/main/java/com/hillel/items_exchange/service/AdvertisementService.java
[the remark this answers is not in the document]
rpkyrych: @lubchik-14 Thank you for paying attention here.

PR #86, opened 2020-07-22
CGE-93: Created image compressing

rpkyrych:


src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -56,22 +59,24 @@
...
+        boolean isAllSupportedTypesReceived = images.parallelStream()
                 .map(MultipartFile::getContentType)
                 .allMatch(SUPPORTED_TYPES::contains);
         if (!isAllSupportedTypesReceived) {
             return new ResponseEntity<>(HttpStatus.UNSUPPORTED_MEDIA_TYPE);
         }
 
         try {
-            imageService.saveToProduct(productId, photos);
+            Product productToSaveImages = productService.findById(productId).orElseThrow(ClassNotFoundException::new);
```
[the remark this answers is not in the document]
rpkyrych: EntityNotFoundException is RuntimeException and we have ExceptionHandler for further working with such exception that could be thrown unexpected. ClassNotFound is checked exception and I catch it later here. So I'll keep it as is.

src/main/java/com/hillel/items_exchange/service/impl/ImageServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+            compressedImages.add(compressImage(photo));
+        }
+        return compressedImages;
+    }
+
+    @Override
+    public byte[] compressImage(MultipartFile image) throws IOException {
+
+        String contentType = image.getContentType();
+        if (contentType == null || contentType.equals(MediaType.IMAGE_GIF_VALUE)) {
```
lubchik-14, line 65: I noticed you check in all content types of received images in Image controller at 66-68 lines, don't you? Why do you check content type again?
Also you don't compress gif files at all, am I right? Why?

src/test/java/com/hillel/items_exchange/controller/ImageControllerIntegrationTest.java
```diff
@@ -69,16 +67,6 @@ void getImageLinksByProductId_shouldReturnAllImageLinks() throws Exception {
...
-    @WithMockUser("admin")
-    @Test
-    @ExpectedDataSet(value = "image/create.yml", ignoreCols = {"created", "updated"})
-    void saveImages_shouldWriteNewImageWhenSupportedType() throws Exception {
-        mockMvc.perform(multipart("/image/{product_id}", 2L)
-                .file(jpeg).file(png))
-                .andDo(print())
-                .andExpect(status().isOk());
-    }
-
```
lubchik-14, line 81: Please, could you explain why you deleted this test case?

src/main/java/com/hillel/items_exchange/service/ImageService.java
```diff
@@ -1,58 +1,70 @@
...
+     */
+    List<byte[]> compressImages(List<MultipartFile> images) throws IOException;
+
+    /**
+     * Make in-memory compressing (20% of basic quality) only for supported types of images
+     * @param image image for further compression
+     * @return compresses image's bytes
+     * @throws IOException in cases when received image is corrupted or it's impossible to read it properly
+     */
+    byte[] compressImage(MultipartFile image) throws IOException;
```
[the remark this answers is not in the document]
rpkyrych: I agree - will change.

src/main/java/com/hillel/items_exchange/service/impl/ImageServiceImpl.java
rpkyrych: @lubchik-14 .gif files couldn't be compressed at all. I agree that it worth to check file types here as well - I'll do it.

src/test/java/com/hillel/items_exchange/controller/ImageControllerIntegrationTest.java
rpkyrych: Mocked controller couldn't work with real images. So I had to remove test for that here and to create new ones into ImageServiceTest.

src/main/java/com/hillel/items_exchange/service/impl/ImageServiceImpl.java
```diff
@@ -115,4 +121,18 @@ public void removeById(List<Long> imageIdList) {
...
+        }
+    }
+
+    private Optional<String> findUnsupportedType(List<MultipartFile> images) {
+        return images.parallelStream()
+                .map(MultipartFile::getContentType)
+                .filter(s -> !supportedTypes.contains(s))
+                .findAny();
+    }
 }
```
lubchik-14: In my opinion better to return all of given unsuppurted types or only boolean value instead of returning any one.
Also I can't test but I think it is wasteful way to make parallel stream from collection of 10 elements.

rpkyrych: Done

PR #89, opened 2020-07-24
CGE-112 Fixed authentication bug

drw85:
Fixed authentication bug.
BCryptPasswordEncoder coding result every time is different because of salt. So it was problem in comparing of stored in db password and result of BCryptPasswordEncoder coding in straight. Using method 'matches' of BCryptPasswordEncoder solved the problem.

### DIFF
--- a/src/main/java/com/hillel/items_exchange/service/UserService.java
+++ b/src/main/java/com/hillel/items_exchange/service/UserService.java
@@ -6,6 +6,7 @@ import com.hillel.items_exchange.dto.UserRegistrationDto;
 import com.hillel.items_exchange.mapper.UserMapper;
 import com.hillel.items_exchange.model.Role;
 import com.hillel.items_exchange.model.User;
+import com.hillel.items_exchange.util.PatternHandler;
 import lombok.RequiredArgsConstructor;
 import org.modelmapper.ModelMapper;
 import org.springframework.beans.BeanUtils;
@@ -13,6 +14,7 @@ import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.stereotype.Service;
 
 import java.util.Optional;
+import java.util.regex.Pattern;
 
 @Service
 @RequiredArgsConstructor
@@ -27,7 +29,11 @@ public class UserService {
     }
 
     public boolean existsByUsernameOrEmailAndPassword(String usernameOrEmail, String encryptedPassword) {
-        return existsByUsername(usernameOrEmail) || existsByEmailAndPassword(usernameOrEmail, encryptedPassword);
+        Pattern emailPattern = Pattern.compile(PatternHandler.EMAIL);
+        if (emailPattern.matcher(usernameOrEmail).matches()) {
+            return existsByEmailAndPassword(usernameOrEmail, encryptedPassword);
+        }
+        return existsByUsernameAndPassword(usernameOrEmail, encryptedPassword);
     }
 
     public boolean registerNewUser(UserRegistrationDto userRegistrationDto, Role role) {
@@ -51,10 +57,13 @@ public class UserService {
     }
 
     public boolean existsByEmailAndPassword(String email, String encodedPassword){
-      return userRepository.findAll().stream()
-              .anyMatch(user ->
-                      user.getEmail().equals(email)
-                              && bCryptPasswordEncoder.matches(encodedPassword, user.getPassword()));
+        final Optional<User> user = userRepository.findByEmail(email);
+        return user.filter(u -> bCryptPasswordEncoder.matches(encodedPassword, u.getPassword())).isPresent();
+    }
+
+    public boolean existsByUsernameAndPassword(String username, String encodedPassword){
+        final Optional<User> user = userRepository.findByUsername(username);
+        return user.filter(u -> bCryptPasswordEncoder.matches(encodedPassword, u.getPassword())).isPresent();
     }
 
     public Optional<UserDto> getByUsernameOrEmail(String usernameOrEmail) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -27,7 +27,7 @@
     }
 
     public boolean existsByUsernameOrEmailAndPassword(String usernameOrEmail, String encryptedPassword) {
-        return userRepository.existsByUsernameOrEmailAndPassword(usernameOrEmail, usernameOrEmail, encryptedPassword);
+        return existsByUsername(usernameOrEmail) || existsByEmailAndPassword(usernameOrEmail, encryptedPassword);
```
lubchik-14, line 30: This method returns true if the first condition is true, so it doesn't check in this case password at all, doesn't it?

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -50,11 +50,20 @@ public boolean existsByEmail(String email) {
...
+    }
+
     public Optional<UserDto> getByUsernameOrEmail(String usernameOrEmail) {
         return userRepository.findByUsername(usernameOrEmail).map(this::mapUserToDto);
     }
 
     private UserDto mapUserToDto(User user) {
         return modelMapper.map(user, UserDto.class);
     }
+
```
lubchik-14, line 67: Remove these two empty lines, please.

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -50,11 +50,20 @@ public boolean existsByEmail(String email) {
         return userRepository.existsByEmail(email);
     }
 
+    public boolean existsByEmailAndPassword(String email, String encodedPassword){
+      return userRepository.findAll().stream()
```
lubchik-14, line 54: Why do you take all users from db instead of talking one by email?

src/main/java/com/hillel/items_exchange/service/UserService.java
drw85: Yes. Parent method wants to know is any of conditions is true.

src/main/java/com/hillel/items_exchange/service/UserService.java
drw85: My bad) Will be removed.

src/main/java/com/hillel/items_exchange/service/UserService.java
drw85: Agree. Your way is better. Will be changed.

src/main/java/com/hillel/items_exchange/service/UserService.java
rpkyrych: Your method lost primary behavior - so if the user exist it doesn't check it's password. I already told you that you should check @ symbol into the gained usernameOrEmail attribute and invoke existsByUsernameAndPasword or existsByEmailAndPassword

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -50,6 +56,16 @@ public boolean existsByEmail(String email) {
...
     }
 
+    public boolean existsByEmailAndPassword(String email, String encodedPassword){
+        final Optional<User> user = userRepository.findByEmail(email);
+        return user.filter(u -> bCryptPasswordEncoder.matches(encodedPassword, u.getPassword())).isPresent();
+    }
+
+    public boolean existsByUsernameAndPassword(String username, String encodedPassword){
+        final Optional<User> user = userRepository.findByUsername(username);
+        return user.filter(u -> bCryptPasswordEncoder.matches(encodedPassword, u.getPassword())).isPresent();
```
rpkyrych, line 66: Extract duplicate code to a new method

drw85: Done

src/main/java/com/hillel/items_exchange/service/UserService.java
lubchik-14: @rpkyrych I don't understand why creating two new methods `existsByUsernameAndPasword` and `existsByEmailAndPassword` is better than 
```
findByUsernameOrEmail(usernameOrEmail)
                .filter(u -> isPasswordMatches(u, encryptedPassword))
                .isPresent();
```

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -27,7 +29,14 @@
     }
 
     public boolean existsByUsernameOrEmailAndPassword(String usernameOrEmail, String encryptedPassword) {
-        return userRepository.existsByUsernameOrEmailAndPassword(usernameOrEmail, usernameOrEmail, encryptedPassword);
+        Pattern emailPattern = Pattern.compile(PatternHandler.EMAIL);
+        Optional<User> user;
```
rpkyrych, line 33: You could simplify it in such way:
```
Optional<User> user = emailPattern.matcher(usernameOrEmail).matches()
                ? userRepository.findByEmail(usernameOrEmail)
                : userRepository.findByUsername(usernameOrEmail);
```

src/test/java/com/hillel/items_exchange/util/AuthControllerIntegrationTestUtil.java
```diff
@@ -20,4 +21,8 @@ protected UserRegistrationDto createUserRegistrationDto(String username, String
                                                             String confirmPassword) {
         return new UserRegistrationDto(username, email, password, confirmPassword);
     }
+
+    protected UserLoginDto createUserAuthenticationDto(String username, String password) {
```
rpkyrych, line 25: From my point of view such method is excess because it's used only once.

PR #96, opened 2020-08-18
CGE-111 BugFix Email validation when registering new user

drw85:
According to RFC 5321 specification the format of email addresses is **local-part**@**domain** where the local part may be up to 64 octets long and the domain may have a maximum of 255 octets. (4.5.3.1.1.  Local-part, 4.5.3.1.2.  Domain).
Was changed:
- refactored exception handling of email validation


Jack11M (approved):

drw85:
Will be created new one

PR #99, opened 2020-08-22
CGE-105 Create CRUD endpoint for Location

lubchik-14:
Created REST endpoint '/location' that will provide access to CRUD operation with it.
Only registered users could use this endpoint.
New code is covered by unit tests.

### DIFF
--- a/src/main/java/com/hillel/items_exchange/controller/LocationController.java
+++ b/src/main/java/com/hillel/items_exchange/controller/LocationController.java
@@ -49,8 +49,7 @@ public class LocationController {
     public ResponseEntity<LocationDto> createLocation(@Valid @RequestBody LocationDto locationDto) {
         long id = locationDto.getId();
         if (id != 0) {
-            String message = String.format("Location not created. Received ID=%s", id);
-            log.warn(message);
+            log.warn("Location not created. Received ID={}", id);
             throw new IllegalIdentifierException(getExceptionMessageSourceWithId(id, "new.location.id.not-zero"));
         }
         return new ResponseEntity<>(locationService.save(locationDto), HttpStatus.CREATED);
@@ -76,7 +75,7 @@ public class LocationController {
             String strIds = locationIds.stream()
                     .map(String::valueOf)
                     .collect(Collectors.joining(", ", ": ", ""));
-            log.warn("Received nonexistent IDs" + strIds);
+            log.warn("Received nonexistent IDs {}", strIds);
             throw new IllegalIdentifierException(
                     getExceptionMessageSourceWithAdditionalInfo("exception.illegal.id", strIds));
         }
--- a/src/main/java/com/hillel/items_exchange/controller/LocationController.java
+++ b/src/main/java/com/hillel/items_exchange/controller/LocationController.java
@@ -1,0 +1,75 @@
+package com.hillel.items_exchange.controller;
+
+import com.hillel.items_exchange.dto.LocationDto;
+import com.hillel.items_exchange.service.LocationService;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.hibernate.boot.model.naming.IllegalIdentifierException;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.*;
+
+import javax.validation.Valid;
+import javax.validation.constraints.PositiveOrZero;
+import java.util.List;
+
+import static com.hillel.items_exchange.config.SecurityConfig.HAS_ROLE_ADMIN;
+import static com.hillel.items_exchange.util.MessageSourceUtil.getExceptionMessageSource;
+import static com.hillel.items_exchange.util.MessageSourceUtil.getExceptionMessageSourceWithId;
+
+@RestController
+@RequestMapping("/location")
+@RequiredArgsConstructor
+@Validated
+@Slf4j
+public class LocationController {
+    private final LocationService locationService;
+
+    @GetMapping
+    public ResponseEntity<List<LocationDto>> getAllLocations() {
+        return new ResponseEntity<>(locationService.findAll(), HttpStatus.OK);
+    }
+
+    @GetMapping("/{location_id}")
+    public ResponseEntity<LocationDto> getLocation(@PathVariable("location_id")
+                                                   @PositiveOrZero(message = "{invalid.id}") long id) {
+        try {
+            LocationDto locationDto = locationService.getById(id);
+            return new ResponseEntity<>(locationDto, HttpStatus.OK);
+        } catch (ClassNotFoundException e) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+    }
+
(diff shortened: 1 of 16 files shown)


src/main/java/com/hillel/items_exchange/service/impl/LocationServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+import static com.hillel.items_exchange.util.MessageSourceUtil.getExceptionMessageSource;
+
+@Service
+@RequiredArgsConstructor
+public class LocationServiceImpl implements LocationService {
+    private final LocationRepository locationRepository;
+
+    @Override
+    public List<LocationDto> findAll() {
+        return new ArrayList<>(convertAllTo(locationRepository.findAll(), LocationDto.class, ArrayList::new));
```
Jack11M, line 25: Great approach!

Jack11M (approved):
Looks good!

src/main/java/com/hillel/items_exchange/service/LocationService.java
```diff
@@ -1,22 +1,70 @@
...
-        return locationRepository.save(location);
-    }
-}
+    /**
+     * If a location exists, returns the Location DTO, otherwise throws {@code ClassNotFoundException}.
+     * @param id Location ID.
+     * @return the non-{@code null} {@link LocationDto} with given id.
+     * @throws ClassNotFoundException if no location found.
+     */
+    LocationDto getById(long id) throws ClassNotFoundException;
```
rpkyrych, line 29: Why don't you use Optional here?

src/main/java/com/hillel/items_exchange/service/LocationService.java
```diff
@@ -1,22 +1,70 @@
...
+    Optional<Location> findById(long id);
 
-    public Location createLocation(Location location) {
-        return locationRepository.save(location);
-    }
-}
+    /**
+     * If a location exists, returns the Location DTO, otherwise throws {@code ClassNotFoundException}.
+     * @param id Location ID.
+     * @return the non-{@code null} {@link LocationDto} with given id.
```
rpkyrych, line 26: :cupid:

src/main/java/com/hillel/items_exchange/service/impl/LocationServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public Optional<Location> findById(long id) {
+        return locationRepository.findById(id);
+    }
+
+    @Override
+    public LocationDto getById(long id) throws ClassNotFoundException {
+        return findById(id)
+                .map(location -> convertTo(location, LocationDto.class))
+                .orElseThrow(() -> new ClassNotFoundException(getExceptionMessageSource("exception.illegal.id")));
```
rpkyrych, line 37: Why don't you return Optional.empty for such case instead of throwing the exception?

src/main/java/com/hillel/items_exchange/controller/LocationController.java
```diff
@@ -0,0 +1,75 @@
...
+        if (!locationService.existsById(locationDto.getId())) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        }
+        return new ResponseEntity<>(locationService.update(locationDto), HttpStatus.OK);
+    }
+
+    @PreAuthorize(HAS_ROLE_ADMIN)
+    @DeleteMapping
+    public ResponseEntity<HttpStatus> deleteLocations(@RequestParam("ids") List<Long> locationIds) {
+        boolean isAllMatch = locationIds.stream().allMatch(locationService::existsById);
```
rpkyrych, line 68: It's really bad idea to check every ID because for each ID will be created SQL request. It's better to get all locations and to check ID there. You could create a new method on LocationService that will get list of id and return boolean as result.

src/main/resources/liquibase/db.changelog-1.0.xml
```diff
@@ -446,4 +446,17 @@
...
     </changeSet>
+
+    <changeSet id="1.0.27: SET NULL constraint update onDelete in advertisement table" author="lubchik">
+        <dropForeignKeyConstraint baseTableName="advertisement"
+                                  constraintName="fk_advertisement_to_location"/>
+
+        <addForeignKeyConstraint baseColumnNames="location_id"
+                                 baseTableName="advertisement"
+                                 constraintName="fk_advertisement_to_location"
+                                 onDelete="SET NULL"
```
rpkyrych, line 457: :+1:

src/test/java/com/hillel/items_exchange/controller/LocationControllerIntegrationTest.java
```diff
@@ -0,0 +1,185 @@
...
+                .andDo(print())
+                .andExpect(status().isNotFound())
+                .andReturn();
+    }
+
+    @WithMockUser(username = "admin", roles = {"ADMIN"})
+    @Test
+    @DataSet("database_init.yml")
+    @ExpectedDataSet("location/update.yml")
+    void updateLocation_shouldReturn200Ok() throws Exception {
```
rpkyrych, line 141: Excess 0 into the test name

rpkyrych (changes requested):
Few thing need to be changed. All other code looks fantastic!

src/main/java/com/hillel/items_exchange/service/LocationService.java
lubchik-14: Because of "get" in the name of the method. I would like to name this method as "find", but I can't until `Optional<Location> findById(long id)` is removed. Also I can't use `protected` to overload the method because of  dealing with interface. Maybe you suggest me something clever at this case?

src/main/java/com/hillel/items_exchange/controller/LocationController.java
lubchik-14: Thanks for solution!

src/main/java/com/hillel/items_exchange/controller/LocationController.java
```diff
@@ -0,0 +1,75 @@
...
+
+import static com.hillel.items_exchange.config.SecurityConfig.HAS_ROLE_ADMIN;
+import static com.hillel.items_exchange.util.MessageSourceUtil.getExceptionMessageSource;
+import static com.hillel.items_exchange.util.MessageSourceUtil.getExceptionMessageSourceWithId;
+
+@RestController
+@RequestMapping("/location")
+@RequiredArgsConstructor
+@Validated
+@Slf4j
```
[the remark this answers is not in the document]
lubchik-14: I got

src/main/java/com/hillel/items_exchange/controller/LocationController.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @PreAuthorize(HAS_ROLE_ADMIN)
+    @PostMapping
+    @ApiOperation(value = "Save a new Location. Only for users authorized as admin.")
+    public ResponseEntity<LocationDto> createLocation(@Valid @RequestBody LocationDto locationDto) {
+        long id = locationDto.getId();
+        if (id != 0) {
+            String message = String.format("Location not created. Received ID=%s", id);
+            log.warn(message);
```
rpkyrych, line 53: You could make it in such way: `log.warn("Location not created. Received ID={}", id);` and logger will add ID value into.

rpkyrych (approved):
Great job! :heart:

PR #108, opened 2020-09-27
CGE-118: Add child quantity bug

KozminAlexandr:
Fixed bug with the ability to add more than 10 children. ChildDto now divided to ChildAddDto and ChildUpdateDto. Added end-to-end tests for new functionality

### DIFF
--- a/src/main/java/com/hillel/items_exchange/controller/UserController.java
+++ b/src/main/java/com/hillel/items_exchange/controller/UserController.java
@@ -1,7 +1,9 @@
 package com.hillel.items_exchange.controller;
 
-import com.hillel.items_exchange.dto.ChildDto;
+import com.hillel.items_exchange.dto.ChildAddDto;
+import com.hillel.items_exchange.dto.ChildUpdateDto;
 import com.hillel.items_exchange.dto.UserDto;
+import com.hillel.items_exchange.exception.EntityAmountException;
 import com.hillel.items_exchange.exception.IllegalOperationException;
 import com.hillel.items_exchange.mapper.UtilMapper;
 import com.hillel.items_exchange.model.Child;
@@ -53,20 +55,18 @@ public class UserController {
     }
 
     @GetMapping("/child")
-    public ResponseEntity<List<ChildDto>> getChildren(Principal principal) {
+    public ResponseEntity<List<ChildUpdateDto>> getChildren(Principal principal) {
         return new ResponseEntity<>(userService.getChildren(getUser(principal.getName())), HttpStatus.OK);
     }
 
     @PostMapping("/child")
     public ResponseEntity<HttpStatus> addChildren(@RequestBody
-                                                  @Size(min = 1, message = "{exception.invalid.dto}")
-                                                          List<@Valid ChildDto> childrenDto,
+                                                  @Size(min = 1, max = 10, message = "{exception.invalid.dto}")
+                                                          List<@Valid ChildAddDto> childrenDto,
                                                   Principal principal) {
-        if (childrenDto.stream().anyMatch(dto -> dto.getId() > 0)) {
-            throw new IllegalIdentifierException(
-                    getExceptionMessageSourceWithAdditionalInfo(
-                            "exception.illegal.id",
-                            "Zero ID is expected"));
+        if(childrenDto.size() + userService.getChildren(getUser(principal.getName())).size() >= 10) {
+            throw new EntityAmountException(
+                getExceptionMessageSource("exception.max-amount-of-children"));
         }
         userService.addChildren(getUser(principal.getName()), childrenDto);
         return new ResponseEntity<>(HttpStatus.OK);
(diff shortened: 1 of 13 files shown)


src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -53,20 +55,18 @@
...
-                                                          List<@Valid ChildDto> childrenDto,
+                                                  @Size(min = 1, max = 10, message = "{exception.invalid.dto}")
+                                                          List<@Valid ChildAddDto> childrenDto,
                                                   Principal principal) {
-        if (childrenDto.stream().anyMatch(dto -> dto.getId() > 0)) {
-            throw new IllegalIdentifierException(
-                    getExceptionMessageSourceWithAdditionalInfo(
-                            "exception.illegal.id",
-                            "Zero ID is expected"));
+        if(childrenDto.size() + userService.getChildren(getUser(principal.getName())).size() >= 10) {
```
drw85, line 67: Maybe something like this for better understanding? But it's up to you)))
```suggestion
             int amountOfChildren = childrenDto.size() + userService.getChildren(getUser(principal.getName())).size();
        if (amountOfChildren > 10) {
```

drw85 (approved):

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
```diff
@@ -128,6 +129,15 @@
...
 
+    @ExceptionHandler(EntityAmountException.class)
+    public ResponseEntity<ErrorMessage> handleEntityAmountException(EntityAmountException e,
+                                                                    ServletWebRequest request) {
+        ErrorMessage errorMessage = getErrorMessage(request, HttpStatus.NOT_ACCEPTABLE,
+                "exception.max-amount-of-children", Collections.singletonList(e.getLocalizedMessage()));
+        logErrorMessage(WARN, errorMessage);
+        return new ResponseEntity<>(errorMessage, HttpStatus.NOT_ACCEPTABLE);
+    }
+
```
lubchik-14, line 140: I think you also shoud test this handling in `GlobalExceptionHandlerTest`

lubchik-14 (changes requested):
Cool! But I noticed there is an unexpected behavior fetching children from DB. If records from `children` table from DB has duplicates by `sex` and `birth_date` fields the result `Set` contains only the first element and ignores the others. You can not exclude the `id` to EqualsAndHashCode to avoid add extra children because of it

src/main/java/com/hillel/items_exchange/controller/UserController.java
rpkyrych: @drw85 have you tried to add Size annotation to the DTO and validate children size with it?

drw85: @rpkyrych It will resolve the issue of the incoming size, but we need total amount of user children to be lees than 10. Right?

rpkyrych: So you could use Max(10) to allow values less than allowed.

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -53,20 +55,18 @@
     }
 
     @GetMapping("/child")
-    public ResponseEntity<List<ChildDto>> getChildren(Principal principal) {
+    public ResponseEntity<List<ChildUpdateDto>> getChildren(Principal principal) {
```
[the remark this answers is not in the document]
rpkyrych: @lubchik-14, nice catch!

src/main/java/com/hillel/items_exchange/controller/UserController.java
rpkyrych: I checked it with clean eyes and agree with @drw85 and agreed that it would nice to have additional variable here.

src/main/java/com/hillel/items_exchange/dto/ChildAddDto.java
```diff
@@ -0,0 +1,21 @@
...
+@Getter
+@Setter
+@Builder
+public class ChildAddDto {
+    @NotNull(message = "{invalid.not-null}")
+    private String sex;
+    @PastOrPresent(message = "{invalid.past-or-present.date}")
+    @JsonFormat(pattern = "yyyy-MM-dd")
+    private LocalDate birthDate;
+}
```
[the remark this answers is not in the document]
rpkyrych: Have the same opinion.

rpkyrych (changes requested):
You shouldn't use new DTOs for each CRUD operation. Do it with only one DTO and use interfaces for new and exist objects.

src/main/java/com/hillel/items_exchange/controller/UserController.java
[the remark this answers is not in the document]
KozminAlexandr: Ok, I will fix this

src/main/java/com/hillel/items_exchange/controller/UserController.java
KozminAlexandr: You are all right, I will add additional variable here

src/main/java/com/hillel/items_exchange/dto/ChildAddDto.java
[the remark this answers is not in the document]
KozminAlexandr: I will delete it

src/main/java/com/hillel/items_exchange/exception/EntityAmountException.java
```diff
@@ -0,0 +1,11 @@
+package com.hillel.items_exchange.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(HttpStatus.NOT_ACCEPTABLE)
```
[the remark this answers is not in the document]
KozminAlexandr: Because.. I can't remember why =) I will remove this annotation from here

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
KozminAlexandr: Ok. you are right

PR #112, opened 2020-10-06
CGE-88 Update Personal Information

lubchik-14:
Bug fixes according to the given [checklist](https://docs.google.com/spreadsheets/d/1KpNVqHNJDDaJCVtYKB1jM8Oe3aO7ULuVuL4pguehwJc/edit#gid=0)


src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -112,6 +118,20 @@ private User getUser(String username) {
                 getExceptionMessageSource("exception.user.not-found")));
     }
 
+    private void checkReadOnlyFieldsUpdate(User user, UserDto userDto) throws IllegalOperationException {
```
rpkyrych, line 121: In my oppinion it would be better to check all read-only fields and throw an exception with information about them all, than doing the same but only for one field at the moment.

src/main/java/com/hillel/items_exchange/dto/UserDto.java
```diff
@@ -19,23 +20,29 @@
...
 @Setter
 public class UserDto {
+    @ApiModelProperty(required = true)
     @NotEmpty(message = "{invalid.not-empty}")
     @Size(min = 2, max = 50, message = "{invalid.size}")
     private String username;
+    @ApiModelProperty(required = true)
     @NotEmpty(message = "{invalid.not-empty}")
-    @Size(max = 129, message = "{invalid.email}")
+    @Size(min = 1, max = 129, message = "{invalid.email}")
```
rpkyrych, line 29: Could you eplain why did you do that? We have annotation Email that works over standarts.

src/main/java/com/hillel/items_exchange/util/PatternHandler.java
```diff
@@ -11,5 +11,5 @@
     public static final String PHONE_NUMBER =
             "^\\s*(?<country>\\+?\\d{2})[-. (]*(?<area>\\d{3})[-. )]*(?<number>\\d{3}[-. ]*\\d{2}[-. ]*\\d{2})\\s*$";
     public static final String USERNAME = "(?=\\S+$)[\\wА-Яа-я\\p{Punct}]+";
-    public static final String WORD_EMPTY_OR_MIN_2_MAX_50 = "^$|^[\\w-']{2,50}+$";
+    public static final String WORD_EMPTY_OR_MIN_2_MAX_50 = "^$|^[\\wА-Яа-я-'`ҐЄІЇієїґ]{2,50}+$";
```
rpkyrych, line 14: Great that you've solved such problem! Additionally I'd created a task to chek all other regexp.

rpkyrych (changes requested):
Would you mind making small update according to my comment?

src/main/java/com/hillel/items_exchange/util/PatternHandler.java
svyatoslav145: public static final String WORD_EMPTY_OR_MIN_2_MAX_50 = "^$|^[\\wА-Яа-я-'`ҐЄІЇієїґ]{2,50}+$";
Can I rewrite tests based on new value WORD_EMPTY_OR_MIN_2_MAX_50? 
Can I add "Ё", "ё"?

rpkyrych: Yep, do it, but after current pull-request will be merged.

src/main/java/com/hillel/items_exchange/dto/UserDto.java
lubchik-14: Yes we have, but Swagger by default displays that this field can have min lenth = 0, so I added some constraint to show a requirement of email field

src/main/java/com/hillel/items_exchange/util/PatternHandler.java
GaiverK: I think the plus is useless in this regexp

src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -46,9 +51,13 @@
     public ResponseEntity<UserDto> updateUserInfo(@Valid @RequestBody UserDto userDto, Principal principal)
```
KozminAlexandr, line 51: You can change return type of this method to void, add @ResponceStatus.ACCEPTED annotation and delete return statement

KozminAlexandr (approved):
Looks good! Please, check my message, maybe you will find it useful =)

src/main/java/com/hillel/items_exchange/controller/UserController.java
lubchik-14: I agree with your suggestion about status, but don't about return type. After updating front-end dev can use new data without sending extra  GET requests to get them

GaiverK (approved):

Jack11M (approved):
Great job!!!

rpkyrych (approved):
You've done lots of work! Thanks!

PR #130, opened 2020-11-12
CGE-138: Fixed bug with having more than 10 images into the product, …

rpkyrych:
…created unit tests


src/test/java/com/hillel/items_exchange/end2end/ImageFlowTest.java
```diff
@@ -67,15 +65,6 @@ void getImageLinksByProductId_shouldReturnAllImageLinks() throws Exception {
                 .andExpect(jsonPath("$[1].defaultPhoto").value(true));
     }
 
-    @WithMockUser("admin")
-    @Test
-    void saveImages_shouldReturn400WhenProductIsNotExist() throws Exception {
```
rpkyrych, line 72: Moved to ImageControllerTest, because such class have unit tests only.

src/main/java/com/hillel/items_exchange/controller/ImageController.java
```diff
@@ -78,11 +80,15 @@
             @ApiResponse(code = 406, message = "NOT ACCEPTABLE"),
             @ApiResponse(code = 415, message = "UNSUPPORTED MEDIA TYPE")})
     public ResponseEntity<String> saveImages(@PathVariable("product_id")
-                                                     @PositiveOrZero(message = "{invalid.id}") long productId,
-                                                 @RequestParam(value = "files") @Size(min = 1, max = 10) List<MultipartFile> images) {
+                                             @PositiveOrZero(message = "{invalid.id}") long productId,
+                                             @RequestParam(value = "files") @Size(min = 1, max = 10) List<MultipartFile> images)
+            throws ElementsNumberExceedException {
```
svyatoslav145, line 86: Why didn't you add a constant (something like MAX_IMAGES_AMOUNT) for using it in all place?

rpkyrych: From mine point of view it isn't so critical. I used such value only two time here and each time with different meaning: inside Side annotation as limitation for a request and inside if block as limitation for total quantity of children. 
Let's wait for other guys opinion about extracting it into a new variable or property.

src/main/java/com/hillel/items_exchange/exception/ElementsNumberExceedException.java
```diff
@@ -0,0 +1,11 @@
+package com.hillel.items_exchange.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(HttpStatus.NOT_ACCEPTABLE)
+public class ElementsNumberExceedException extends Exception {
```
KozminAlexandr, line 7: We allready have AntityAmountException for this case (added in CGE-118). Would you like to use it or if you don't like name - just rename it, because they are identical

src/main/java/com/hillel/items_exchange/controller/ImageController.java
KozminAlexandr: Maybe should to extract this variable to global variables? This solution of course not allow to use it in `@Size` annotation, but this will allow us in case we want to change this variable - change it in application.properties and that's all. P.S. you can see how it is realized in CGE-118. But be carefull, I find some unwanted behavior at display message error, I prupose solution from day to day at new PR

KozminAlexandr (changes requested):

src/main/java/com/hillel/items_exchange/controller/ImageController.java
lubchik-14: Our application grows :) and we have some constant values used at different places like max children amount or max images one, so I'm agreed each of them should be gotten from one place. But moving them to the application.properties doesn't allow us to use them with `@Size`, so in my opinion the place to move should be one for all usages from or do nothing with it :)

lubchik-14 (approved):
👍

src/main/java/com/hillel/items_exchange/controller/ImageController.java
rpkyrych: OK. I'll done it only for the method body.

src/main/java/com/hillel/items_exchange/exception/ElementsNumberExceedException.java
rpkyrych: I'll rename it. Thanks.

svyatoslav145 (approved):

Jack11M (approved):
Looks good!

KozminAlexandr (approved):

PR #132, opened 2020-11-28
CGE-118: add child quantity bug

KozminAlexandr:
Added response bodies for /child/post and /child/put. Changed response code for /child/delete from OK to NO_CONTENT


src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -101,20 +101,23 @@ private UserDto mapUserToDto(User user) {
         return convertToDto(parent.getChildren(), ChildDto.class);
     }
 
-    public void addChildren(User parent, List<ChildDto> childrenDtoToAdd) {
+    public List<ChildDto> addChildren(User parent, List<ChildDto> childrenDtoToAdd) {
         final List<Child> childrenToSave = new ArrayList<>(convertAllTo(
                 childrenDtoToAdd, Child.class, ArrayList::new));
         addNewChildren(parent, childrenToSave);
         userRepository.save(parent);
+        return findDuplicatedChildAndConvertToDto(parent.getChildren(), childrenToSave);
```
rpkyrych, line 109: Naming the method in such way you're ruined SOLID principle of Single Responsibility. You have to split such method into two: find duplicates and convert dto. Or delegate mapping to another method and invoke it inside the first one.

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -101,20 +101,23 @@ private UserDto mapUserToDto(User user) {
...
-    public void updateChildren(User parent, List<ChildDto> childrenDtoToUpdate) {
+    public List<ChildDto> updateChildren(User parent, List<ChildDto> childrenDtoToUpdate) {
         parent.getChildren().forEach(pChild -> childrenDtoToUpdate.forEach(uChild -> {
             if (pChild.getId() == uChild.getId()) {
                 BeanUtils.copyProperties(uChild, pChild);
             }
         }));
         userRepository.saveAndFlush(parent);
+        return findDuplicatedChildAndConvertToDto(parent.getChildren(),
+            new ArrayList<>(convertAllTo(childrenDtoToUpdate, Child.class, ArrayList::new)));
```
rpkyrych, line 120: Wow, you method name indicated that it makes convertation but you do it here one more time... Why?

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -158,4 +161,11 @@ private void addNewPhones(User user, Collection<Phone> phones) {
         phones.forEach(phone -> phone.setUser(user));
         user.getPhones().addAll(phones);
     }
+
+    private List<ChildDto> findDuplicatedChildAndConvertToDto(List<Child> fromDB, List<Child> fromRequest) {
+        return fromDB.stream()
+            .filter(fromRequest::contains)
+            .map(child -> modelMapper.map(child, ChildDto.class))
```
rpkyrych, line 168: Split in into few methods according to SOLID.

rpkyrych (changes requested):
Nice but one update needs to be made.

src/main/java/com/hillel/items_exchange/service/UserService.java
KozminAlexandr: First I convert childrenDtoToUpdate to Child.class, after that I find duplicates (because I can't do that with ChilDto) and at last I convert finded childs to dtos for response

src/main/java/com/hillel/items_exchange/service/UserService.java
KozminAlexandr: Okey, I will changed this

Jack11M (approved):

svyatoslav145 (approved):

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -158,4 +162,10 @@ private void addNewPhones(User user, Collection<Phone> phones) {
         phones.forEach(phone -> phone.setUser(user));
         user.getPhones().addAll(phones);
     }
+
+    private List<Child> findDuplicatedChildren(List<Child> fromDB, List<Child> fromRequest) {
```
rpkyrych, line 166: You don't need such method. ```Collections.retainAll()``` makes exactly the same.

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -101,20 +101,24 @@ private UserDto mapUserToDto(User user) {
         return convertToDto(parent.getChildren(), ChildDto.class);
     }
 
-    public void addChildren(User parent, List<ChildDto> childrenDtoToAdd) {
+    public List<ChildDto> addChildren(User parent, List<ChildDto> childrenDtoToAdd) {
         final List<Child> childrenToSave = new ArrayList<>(convertAllTo(
                 childrenDtoToAdd, Child.class, ArrayList::new));
         addNewChildren(parent, childrenToSave);
         userRepository.save(parent);
+        return convertToDto(findDuplicatedChildren(parent.getChildren(), childrenToSave), ChildDto.class);
```
rpkyrych, line 109: You could do it without extra methods and much more simpler:
```
        List<Child> children = parent.getChildren();
        children.retainAll(childrenToSave);
        return convertToDto(children, ChildDto.class);
```

src/main/java/com/hillel/items_exchange/service/UserService.java
```diff
@@ -101,20 +101,24 @@ private UserDto mapUserToDto(User user) {
...
 
-    public void updateChildren(User parent, List<ChildDto> childrenDtoToUpdate) {
+    public List<ChildDto> updateChildren(User parent, List<ChildDto> childrenDtoToUpdate) {
         parent.getChildren().forEach(pChild -> childrenDtoToUpdate.forEach(uChild -> {
             if (pChild.getId() == uChild.getId()) {
                 BeanUtils.copyProperties(uChild, pChild);
             }
         }));
         userRepository.saveAndFlush(parent);
+        final List<Child> childrenToUpdate = new ArrayList<>(convertAllTo(
```
rpkyrych, line 119: The code above already performs an operation with `childrenDtoToUpdate` so why don't you just to all condition which would collect the same id and after get from user such updated children instead of making another convert operation here?

rpkyrych (changes requested):
Almost done. One small change needs to be implemented.

src/main/java/com/hillel/items_exchange/service/UserService.java
KozminAlexandr: You totally right =)

src/main/java/com/hillel/items_exchange/service/UserService.java
KozminAlexandr: Performed

src/main/java/com/hillel/items_exchange/service/UserService.java
KozminAlexandr: I'm not sure I fully understand you, but I think my solution will work fine =)

rpkyrych (approved):
Awsome!

PR #145, opened 2020-12-28
СGE-172 Bug with empty messages on registering user with not valid credentials

drw85:
Checked all cases with invalid credentials for registering new user.
Added message on success user registering.
Covered with tests.


src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
```diff
@@ -67,7 +68,9 @@ public ErrorMessage handleRuntimeException(RuntimeException e, ServletWebRequest
     }
 
     private ErrorMessage logAndGetErrorMessage(ServletWebRequest request, Exception e, Level level) {
-
+        if (e.getClass().equals(UndeclaredThrowableException.class)){
```
rpkyrych, line 71: `if (e instanceof UndeclaredThrowableException){` looks like more elegant

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
```diff
@@ -67,7 +68,9 @@ public ErrorMessage handleRuntimeException(RuntimeException e, ServletWebRequest
...
+            e = (Exception) ((UndeclaredThrowableException) e).getUndeclaredThrowable();
```
rpkyrych, line 72: Could you explain who throws such exception?

src/main/resources/messages_ru.properties
```diff
@@ -32,6 +32,7 @@ subcategory.not-deletable=Подкатегория не может быть уд
 
 user.not-owner=Пользователь не является владельцем данного объявления: 
 user.not-found=Пользователь не найден, username/email: 
+user.created=Создан новый пользователь:{username}
```
rpkyrych, line 35: I observe extra spaces into English message text but they aren't here. Make both messages more consistent.

src/test/resources/messages_ru.properties
```diff
@@ -32,6 +32,7 @@ subcategory.not-deletable=Подкатегория не может быть уд
 
 user.not-owner=Пользователь не является владельцем данного объявления: 
 user.not-found=Пользователь не найден, username/email: 
+user.created=Создан новый пользователь:{username}
```
rpkyrych, line 35: The same about the message here.

rpkyrych (changes requested):
Small changes need to be provided.

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
drw85: Agree

src/main/resources/messages_ru.properties
drw85: Done

src/test/resources/messages_ru.properties
drw85: Done

src/main/java/com/hillel/items_exchange/exception/handler/GlobalExceptionHandler.java
drw85: It's 'validate' method in UserRegistrationDtoValidator thru @SneakyThrows annotation.

rpkyrych (approved):

svyatoslav145 (approved):

Jack11M (approved):

PR #151, opened 2021-01-13
CGE-121_Implement_AdvertisementTitleDto

drw85:
- Created new class AdvertisementTitleDto;
- Created new endpoint '/adv/thumbnail' to retrieve such dtos;

### DIFF
--- a/src/main/java/com/hillel/items_exchange/dto/AdvertisementTitleDto.java
+++ b/src/main/java/com/hillel/items_exchange/dto/AdvertisementTitleDto.java
@@ -1,0 +1,19 @@
+package com.hillel.items_exchange.dto;
+
+import com.hillel.items_exchange.model.Location;
+import lombok.*;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@EqualsAndHashCode
+public class AdvertisementTitleDto {
+    private long advertisementId;
+    private byte[] image;
+    private String title;
+    private Location location;
+    private String ownerName;
+    private byte[] ownerAvatar;
+}
--- a/src/main/java/com/hillel/items_exchange/dto/AdvertisementTitleDto.java
+++ b/src/main/java/com/hillel/items_exchange/dto/AdvertisementTitleDto.java
@@ -1,6 +1,5 @@
 package com.hillel.items_exchange.dto;
 
-import com.hillel.items_exchange.model.Location;
 import lombok.*;
 
 @Getter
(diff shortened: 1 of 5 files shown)


src/main/java/com/hillel/items_exchange/dto/AdvertisementTitleDto.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.hillel.items_exchange.model.Location;
+import lombok.*;
+
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+@EqualsAndHashCode
```
rpkyrych, line 11: You should exclude **advertisementId** from equals and hashcode

src/main/java/com/hillel/items_exchange/service/impl/AdvertisementServiceImpl.java
```diff
@@ -149,6 +157,19 @@ public void setDefaultImage(Advertisement advertisement, Long imageId, User owne
         }.getType());
     }
 
+    private List<AdvertisementTitleDto> mapAdvertisementsToTitleDto(Collection<Advertisement> advertisements) {
+        return advertisements.stream().map(advertisement ->
+                AdvertisementTitleDto.builder()
```
rpkyrych, line 162: You should extract builder into another method **buildAdvertisementTitle** or something similar to and use is here with link to the method ::

rpkyrych: Or you could use **  BeanUtils.copyProperties** but you all the fields names should be the same

src/test/java/com/hillel/items_exchange/controller/AdvertisementControllerTest.java
```diff
@@ -45,6 +45,16 @@ void findPaginated_shouldBeThrownValidationException() throws Exception {
         assertTrue(mvcResult.getResponse().getContentAsString().contains("must be greater than or equal to 0"));
     }
 
+    @Test
+    void findPaginatedAsThumbnails_shouldBeThrownValidationException() throws Exception {
```
rpkyrych, line 49: We already have negative test for pagination. Would you mind to create positive test instead and validate the fields?

rpkyrych (changes requested):
Looks great but few things needs to be changed.

src/main/java/com/hillel/items_exchange/dto/AdvertisementTitleDto.java
drw85: Yeah. You are right.

src/main/java/com/hillel/items_exchange/service/impl/AdvertisementServiceImpl.java
drw85: Good point. Thnx.

src/test/java/com/hillel/items_exchange/controller/AdvertisementControllerTest.java
drw85: Agree. It must be done. I'll do it.

rpkyrych (approved):

Jack11M (approved):

drw85:
- Created new class AdvertisementTitleDto;
- Created new endpoint '/adv/thumbnail' to retrieve such dtos;
- Covered with tests;

PR #154, opened 2021-01-17
Cge 166 db adv entity update

Jack11M:
Added String values to the AgeRange, renamed FROM_13_TO_14 to FROM_12_TO_14, fixed tests.
Renamed Lang to I18n, fixed tests.

### DIFF
--- a/src/main/java/com/hillel/items_exchange/model/Location.java
+++ b/src/main/java/com/hillel/items_exchange/model/Location.java
@@ -22,7 +22,6 @@ public class Location {
     private String city;
 
     @Enumerated(EnumType.STRING)
-    @Column(name = "i18n")
     private I18n i18n;
 
     @OneToMany(mappedBy = "location")
--- a/src/main/java/com/hillel/items_exchange/model/Location.java
+++ b/src/main/java/com/hillel/items_exchange/model/Location.java
@@ -4,7 +4,7 @@ import lombok.*;
 
 import javax.persistence.*;
 import java.util.List;
-import com.hillel.items_exchange.model.enums.Lang;
+import com.hillel.items_exchange.model.enums.I18n;
 
 @Entity
 @NoArgsConstructor
@@ -23,7 +23,7 @@ public class Location {
 
     @Enumerated(EnumType.STRING)
     @Column(name = "lang")
-    private Lang lang;
+    private I18n i18N;
 
     @OneToMany(mappedBy = "location")
     private List<Advertisement> advertisements;
--- a/src/main/java/com/hillel/items_exchange/model/Location.java
+++ b/src/main/java/com/hillel/items_exchange/model/Location.java
@@ -22,8 +22,8 @@ public class Location {
     private String city;
 
     @Enumerated(EnumType.STRING)
-    @Column(name = "lang")
-    private I18n i18N;
+    @Column(name = "i18n")
+    private I18n i18n;
 
     @OneToMany(mappedBy = "location")
     private List<Advertisement> advertisements;
--- a/src/main/java/com/hillel/items_exchange/model/enums/I18n.java
+++ b/src/main/java/com/hillel/items_exchange/model/enums/I18n.java
@@ -3,14 +3,14 @@ package com.hillel.items_exchange.model.enums;
 import lombok.Getter;
 
 @Getter
-public enum Lang {
-    UA("ua"),
-    EN("en"),
-    RU("ru");
+public enum I18n {
+    UA("UA_uk"),
+    EN("EN_en"),
+    RU("RU_ru");
 
     private final String value;
 
-    Lang(String value) {
+    I18n(String value) {
         this.value = value;
     }
 }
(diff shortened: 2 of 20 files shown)


drw85 (approved):
Looking good. I have one question but I approve.

src/main/java/com/hillel/items_exchange/model/Location.java
```diff
@@ -23,7 +23,7 @@
 
     @Enumerated(EnumType.STRING)
     @Column(name = "lang")
```
rpkyrych, line 25: It's looks like you missed rename it into Column annotation.

src/main/java/com/hillel/items_exchange/model/enums/I18n.java
```diff
@@ -3,14 +3,14 @@
 import lombok.Getter;
 
 @Getter
-public enum Lang {
-    UA("ua"),
-    EN("en"),
-    RU("ru");
+public enum I18n {
+    UA("UA_uk"),
+    EN("EN_en"),
```
rpkyrych, line 8: Awesome! 🥇

rpkyrych (changes requested):
Fix Column and check Sergiy's question.

src/main/java/com/hillel/items_exchange/model/Location.java
Jack11M: No, I have not changed the column name in the DB. We did not discuss this. Is it critical? I can do it, but you will need to remove the DB one more time.

src/test/java/com/hillel/items_exchange/service/LocationServiceTest.java
```diff
@@ -60,7 +59,7 @@ void findAll_shouldReturnAllLocationDtos() {
     void findById_shouldReturnLocation() {
         when(locationRepository.findById(anyLong())).thenReturn(Optional.of(location));
 
-        Location foundLocation = locationService.findById(1L).get();
+        Location foundLocation = locationService.findById(1L).orElseGet(Location::new);
```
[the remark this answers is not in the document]
Jack11M: As I know, it is a really bad solution to use Optional.get() directly without checking ifPresent. At these places we had such approach, so I decided to fix them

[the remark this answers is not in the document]
rpkyrych: You didn't understand the question: you used Mockito above with returning location from there if anyone triggers **.findById** method and here you make such triggering instead the test. So why do we need doing that?

src/main/java/com/hillel/items_exchange/model/Location.java
rpkyrych: Difference between names creates misunderstanding - you should update it or rename field name to lang.

Jack11M: Ok

src/test/java/com/hillel/items_exchange/service/LocationServiceTest.java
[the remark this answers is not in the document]
Jack11M: I understood it. But findById returns Optional! And we just use get method, it is ugly in my opinion.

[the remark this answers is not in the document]
Jack11M: But I can return the previous code of you wish.
By the way, you told me that use Optional.get() unacceptable.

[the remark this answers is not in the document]
Jack11M: Fixed

src/main/java/com/hillel/items_exchange/model/Location.java
Jack11M: Done

src/test/java/com/hillel/items_exchange/service/LocationServiceTest.java
```diff
@@ -72,7 +71,7 @@ void findById_shouldReturnLocation() {
     void getById_shouldReturnLocationDto() {
         when(locationRepository.findById(anyLong())).thenReturn(Optional.of(location));
 
-        LocationDto foundLocation = locationService.getById(1L).get();
+        LocationDto foundLocation = locationService.getById(1L).orElseGet(LocationDto::new);
```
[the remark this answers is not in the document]
Jack11M: Fixed

src/main/java/com/hillel/items_exchange/model/Location.java
```diff
@@ -22,8 +22,8 @@
     private String city;
 
     @Enumerated(EnumType.STRING)
-    @Column(name = "lang")
-    private Lang lang;
+    @Column(name = "i18n")
```
rpkyrych, line 25: You could not to use annotation Column if the names of the column in DB and into entity are absolutely equal

rpkyrych (approved):

src/main/java/com/hillel/items_exchange/model/Location.java
Jack11M: Fixed

PR #155, opened 2021-01-20
CGE-176:Create SomethingBad Page

vss1502:
Create SomethingBad Page


src/main/resources/react/src/components/local/en.js
```diff
@@ -49,6 +49,11 @@ const en = {
 		noPage: 'The page does not exist!',
 		mainPage: 'To the main',
 		backPage: 'Back'
+	},
+	somethingBad: {
+		error: 'Error! What do we do?',
```
rpkyrych, line 54: Wrong translation - needs to be: "What should we do the next?"

rpkyrych (commented):
Would you mind updating the message on English?

Jack11M (approved):
Change the phrase "What do we do next?" to "What do we do next?" (as Roman said) please.

Wolshebnik (approved):

PR #172, opened 2021-03-05
CGE-193 fix bug when it is possible to enter more than 129 character on E-mail fields when changing one

svyatoslav145:


src/main/java/com/hillel/items_exchange/controller/UserController.java
```diff
@@ -101,8 +102,8 @@ public String updateUserPassword(@Valid @RequestBody UserChangePasswordDto userC
             @ApiResponse(code = 403, message = "FORBIDDEN"),
             @ApiResponse(code = 409, message = "CONFLICT")})
     @ResponseStatus(HttpStatus.ACCEPTED)
-    public String updateUserEmail(@Valid @RequestBody UserChangeEmailDto userChangeEmailDto, Principal principal)
-            throws DataConflictException {
+    public String updateUserEmail(@Valid @RequestBody UserChangeEmailDto userChangeEmailDto, BindingResult bindingResult,
```
rpkyrych, line 105: Could you explain why do we need BindingResult here?

src/main/java/com/hillel/items_exchange/dto/UserChangeEmailDto.java
```diff
@@ -18,10 +19,10 @@
 public class UserChangeEmailDto {
 
     @NotEmpty(message = "{invalid.not-empty}")
+    @Size(max = 129, message = "{too.big.email}")
```
rpkyrych, line 22: If I remember correctly we already discussed it and desided just to use Email annotation? @Jack11M , @drw85 do you remember such discussion?

src/main/java/com/hillel/items_exchange/controller/UserController.java
svyatoslav145: I'd like to discuss this.

src/main/java/com/hillel/items_exchange/dto/UserChangeEmailDto.java
svyatoslav145: I'd like to discuss this.

PR #192, opened 2021-04-14
CGE 50 product page fix

SergeyCheremisin:


Wolshebnik (approved):

src/main/resources/react/src/components/local/en.js
```diff
@@ -50,7 +50,8 @@ const en = {
 		addField: 'Add field',
 		removeField: 'Remove field',
 		saveChanges: 'Save changes',
-		enter: 'Enter'
+		enter: 'Enter',
+		look: 'To view',
 	},
```
Jack11M, line 55: I would prefer to change 'to view' to 'View'

src/main/resources/react/src/components/local/en.js
```diff
@@ -93,7 +94,8 @@ const en = {
 		age: 'Age',
 		season: 'Season',
 		sex: 'Sex',
-		button: 'Offer an exchange'
+		button: 'Offer an exchange',
```
Jack11M, line 97: As for me, it is better to use 'exchange offers' or 'offers for exchange' here

src/main/resources/react/src/components/local/en.js
```diff
@@ -50,7 +50,8 @@ const en = {
 		addField: 'Add field',
 		removeField: 'Remove field',
 		saveChanges: 'Save changes',
-		enter: 'Enter'
+		enter: 'Enter',
+		look: 'To view',
```
Jack11M, line 54: I would prefer to change to 'View'

src/main/resources/react/src/components/local/en.js
Jack11M, line 97: As for me, it is better to use 'Exchange offers' or 'Offers for exchange' here

Jack11M (approved):

PR #249, opened 2021-08-01
OBMIN-67: Fixed bug when Postgres DB parses SQL

Jack11M:
Added BinaryType to Lob columns to fix bug when Postgres DB parses SQL


Wolshebnik (approved):

rpkyrych (approved):

Basist92 (approved):

Basist92 (commented):
can you check the need for the same changes in the class model.User?

PR #259, opened 2021-09-26
OBMIN-35: done

Wolshebnik:


src/main/resources/react/src/pages/AddGoods/AddGoods.js
```diff
@@ -142,234 +145,214 @@ const AddGoods = () => {
...
-            <CheckBox
-              text="Рассмотрю Ваше предложение"
-              margin="0 0 15px 0"
-              fontSize="14px"
-              checked={false}
-              click={null}
+    <FormHandler
+      onSubmit={handleSubmit}
+      // initialValues={initialValues}
+      // validationSchema={createSchema}
```
rpkyrych, line 156: Do we really need that comments?

src/main/resources/react/src/components/local/ua.js
```diff
@@ -170,6 +170,10 @@ const ua = {
     headline: 'Заголовок оголошення',
     selectCategory: 'Виберіть категорію',
     selectSubcategory: 'Виберіть підкатегорію',
+    exchange: 'Обмін',
+    whatChange: 'На що хочете обмінятися?',
```
Jack11M, line 174: As for me, it would be better to use 'бажаєте' here

src/main/resources/react/src/components/local/ua.js
```diff
@@ -170,6 +170,10 @@ const ua = {
...
+    enterPhrase: 'введіть фразу, а потім натисніть Enter',
+    placeholderChange: 'Введіть на що хочете обмінятися',
```
Jack11M, line 176: As for me, it would be better to use 'бажаєте' here

Jack11M (approved):

src/main/resources/react/src/pages/AddGoods/AddGoods.js
Wolshebnik: yes. I need for future

PR #290, opened 2021-11-14
OBMIN-73: added text area

Wolshebnik:


MaryGoroshko (approved):

Jack11M (approved):

src/main/resources/react/src/components/local/en.js
```diff
@@ -18,6 +18,11 @@ const en = {
     errorTitle: 'Error',
     notEmptyInput: 'At least one field is required.',
     selectCategory: 'Select the category first',
+    addedFile: 'The file you are adding already exists.',
+    pictureSelection: 'Please select a picture with extension',
+    sizeFile: 'Size of your file',
+    selectFile: 'Select a file less than 10 MB.',
+    noSaveMore: 'You cannot save more than 10 files.',
```
rpkyrych, line 25: It might be better to use “image“ word here

rpkyrych (approved):

src/main/resources/react/src/components/local/en.js
Wolshebnik: fixed

PR #457, opened 2023-01-31
OBMIN-200 Added a user restriction on changing the password

demus-n:
Added a restriction to prevent changing the password if the Current password is entered in fields "New password" and "Repeat password" in Settings Personal area + test


src/test/java/space/obminyashka/items_exchange/controller/UserControllerIntegrationTest.java
```diff
@@ -140,6 +140,16 @@ void updateUserPassword_WhenOldPasswordWrong_ShouldThrowInvalidDtoException() th
         assertTrue(mvcResult.getResponse().getContentAsString().contains(getMessageSource("incorrect.password")));
     }
 
+    @Test
+    @WithMockUser(username = "user")
```
rpkyrych, line 144: The interface WithMockUser has its default values:

```
public @interface WithMockUser {
    String value() default "user";

    String username() default "";

    String[] roles() default {"USER"};

    String[] authorities() default {};

    String password() default "password";
```
So, you can omit using the same username by simply using the annotation with default settings.

src/test/java/space/obminyashka/items_exchange/controller/UserControllerIntegrationTest.java
```diff
@@ -140,6 +140,16 @@ void updateUserPassword_WhenOldPasswordWrong_ShouldThrowInvalidDtoException() th
...
+    void updateUserPassword_WhenUserEnteredOldPassword_ShouldThrowIllegalArgumentException() throws Exception {
```
rpkyrych, line 145: Small changes need to be introduced to the method's name: when and should words have to start with a low letter because of the camel case style.

src/test/java/space/obminyashka/items_exchange/controller/UserControllerIntegrationTest.java
```diff
@@ -140,6 +140,16 @@ void updateUserPassword_WhenOldPasswordWrong_ShouldThrowInvalidDtoException() th
...
+        UserChangePasswordDto userChangePasswordDto = new UserChangePasswordDto(CORRECT_OLD_PASSWORD, CORRECT_OLD_PASSWORD, CORRECT_OLD_PASSWORD);
```
rpkyrych, line 146: Just FYI, you can use **var** instead of the class, which could be `var userChangePasswordDto = new UserChangePasswordDto(...) `.
But it's up to you, of course.

rpkyrych (approved):
Looks good

ennnbaev (approved):
Looks great!

PR #460, opened 2023-02-03
OBMIN-221 Fixed bug with several avatars

IllyaZuew:


src/test/java/space/obminyashka/items_exchange/controller/UserControllerIntegrationTest.java
```diff
@@ -200,6 +203,18 @@ void setUserAvatar_whenReceivedBMPImage_shouldThrowUnsupportedMediaTypeException
         sendUriAndGetMvcResult(multipart(new URI(USER_SERVICE_CHANGE_AVATAR)).file(bmp), status().isUnsupportedMediaType());
     }
 
+    @Test
+    @WithMockUser(username = "admin")
+    void setUserAvatar_whenReceivedSeveralImages_shouldThrowBadRequestException() throws Exception {
```
ennnbaev, line 208: I think updateUserAvatar... would be better than setUserAvatar

src/main/java/space/obminyashka/items_exchange/controller/UserController.java
```diff
@@ -181,13 +182,17 @@ public List<ChildDto> updateChildren(@Size(max = MAX_CHILDREN_AMOUNT, message =
...
             @ApiResponse(code = 202, message = "ACCEPTED"),
+            @ApiResponse(code = 400, message = "BAD REQUEST"),
             @ApiResponse(code = 403, message = "FORBIDDEN"),
             @ApiResponse(code = 406, message = "NOT ACCEPTABLE"),
             @ApiResponse(code = 415, message = "UNSUPPORTED MEDIA TYPE")})
     @ResponseStatus(HttpStatus.ACCEPTED)
-    public Map<String, byte[]> updateUserAvatar(@RequestParam MultipartFile image, @ApiIgnore Authentication authentication) {
+    public Map<String, byte[]> updateUserAvatar(@RequestParam(name = "image") MultipartFile[] images, @ApiIgnore Authentication authentication) throws BadRequestException {
+        if(images.length != 1){
+            throw new BadRequestException("Image should be only one");
```
ennnbaev, line 192: In my mind, it would be better if we use getMessageSource()

src/main/java/space/obminyashka/items_exchange/controller/UserController.java
```diff
@@ -181,13 +182,17 @@ public List<ChildDto> updateChildren(@Size(max = MAX_CHILDREN_AMOUNT, message =
...
+        }
         User user = getUser(authentication.getName());
-        byte[] newAvatarImage = imageService.compress(image);
+        byte[] newAvatarImage = imageService.compress(images[0]);
```
ennnbaev, line 195: It looks dangerous

IllyaZuew: If length does not equal 1, the endpoint will return a 400 error, so line 195 always has length which equals 1.

src/main/java/space/obminyashka/items_exchange/controller/UserController.java
IllyaZuew: fixed

ennnbaev (approved):
Nice!

ImAndVlad (approved):

PR #478, opened 2023-02-15
Obmin 240 user is not able to access the functions after account recovery

Sershok:


src/main/java/space/obminyashka/items_exchange/service/impl/UserServiceImpl.java
```diff
@@ -192,7 +194,7 @@ private boolean isDurationMoreThanNumberOfDaysToKeepDeletedUser(User user) {
 
     @Override
     public void makeAccountActiveAgain(User user) {
-        user.setStatus(ACTIVE);
+        roleService.getRole(savedUsersRole.get(user.getUsername())).ifPresent(user::setRole);
```
IllyaZuew, line 197: I think You should set only role ROLE_USER without using map for activating account, but you should check that user has role ROLE_SELF_REMOVING and after that if it is true, activate account by changing role to USER_ROLE

IllyaZuew: or using @PreAuthorize on method in controller for checking role

IllyaZuew (approved):

ennnbaev (approved):
Looks good

src/main/java/space/obminyashka/items_exchange/dao/RoleRepository.java
```diff
@@ -1,14 +1,23 @@
...
 
 @Repository
 public interface RoleRepository extends JpaRepository<Role, UUID> {
 
     Optional<Role> findByName(String name);
+
+    @Transactional
+    @Modifying
+    @Query("update User set role.id =(select id from Role where name='ROLE_USER') where username = :userName")
+    void updateUserRole(String userName);
```
rpkyrych, line 21: 1. The method's name doesn't fit its behavior. I propose to rename it to **setUserRoleToUserByUsername**
2. "username" is the word in lowercase

src/main/java/space/obminyashka/items_exchange/service/RoleService.java
```diff
@@ -18,4 +18,8 @@ public class RoleService {
     public Optional<Role> getRole(String roleName) {
         return roleRepository.findByName(roleName);
     }
+
+    public void updateUserRole(String username) {
```
rpkyrych, line 22: The name of the method doesn't correspond to its behavior. For the current name, it should be the second parameter **role** in it.

src/main/java/space/obminyashka/items_exchange/service/impl/UserServiceImpl.java
```diff
@@ -192,7 +191,7 @@ private boolean isDurationMoreThanNumberOfDaysToKeepDeletedUser(User user) {
 
     @Override
     public void makeAccountActiveAgain(User user) {
-        user.setStatus(ACTIVE);
+        roleService.updateUserRole(user.getUsername());
         userRepository.saveAndFlush(user);
```
rpkyrych, line 195: Here are a couple of comments on the whole method:
1. You don't need to save and flush the user when you update him in **updateUserRole** method.
2. You have to pass fewer privileges to the method. If you need only a username here pass only it, no need to pass the whole user.

src/test/java/space/obminyashka/items_exchange/end2end/UserFlowTest.java
```diff
@@ -168,7 +168,7 @@ void selfDeleteRequest_WhenDataCorrect_Successfully() throws Exception {
     @WithMockUser(username = "deletedUser", roles = "SELF_REMOVING")
     @DataSet(value = {"database_init.yml", "user/deleted_user_init.yml"})
     @ExpectedDataSet(value = "user/deleted_user_restore_expected.yml", orderBy = {"created", "name"},
-            ignoreCols = {"password", "lastOnlineTime", "updated"})
+            ignoreCols = {"password", "lastOnlineTime", "updated","status"})
```
rpkyrych, line 171: formatting

src/test/java/space/obminyashka/items_exchange/service/UserServiceIntegrationTest.java
```diff
@@ -82,12 +79,14 @@ void testUpdateUserEmail_WhenDataCorrect_Successfully() {
...
     @Test
     void testSelfDeleteRequest_WhenDataCorrect_Successfully() {
+
         when(roleService.getRole(anyString())).thenReturn(Optional.of(new Role(UUID.randomUUID(), "ROLE_SELF_REMOVING", List.of())));
 
         userService.selfDeleteRequest(userWithOldPassword);
 
         assertEquals("ROLE_SELF_REMOVING", userWithOldPassword.getRole().getName());
         verify(userRepository).saveAndFlush(userWithOldPassword);
+
```
rpkyrych, line 89: Why do we need to merge empty spaces from the test?

src/test/java/space/obminyashka/items_exchange/service/UserServiceIntegrationTest.java
```diff
@@ -105,14 +104,6 @@ void testPermanentlyDeleteUsers_ShouldDeleteRequiredUsers() {
         }
     }
 
-    @Test
-    void makeAccountActiveAgain_WhenDataCorrect_Successfully() {
```
rpkyrych, line 109: All the test cases have to be created according to the [test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html).
That means that the same logic has to be covered with the unit, integration, end 2 end tests.
You shouldn't remove the test when you have an end 2 end test for the same part of the code but update that integration test as well.
Each test has its own purpose: end 2 end - check the whole flow; integration - work with Spring context and beans; unit - no Spring context and describes only single behavior of a tested method.

rpkyrych (changes requested):

src/test/java/space/obminyashka/items_exchange/service/UserServiceIntegrationTest.java
```diff
@@ -88,8 +85,17 @@ void testSelfDeleteRequest_WhenDataCorrect_Successfully() {
 
         assertEquals("ROLE_SELF_REMOVING", userWithOldPassword.getRole().getName());
         verify(userRepository).saveAndFlush(userWithOldPassword);
+
     }
+    @Test
+    void makeAccountActiveAgain_WhenDataCorrect_Successfully() {
```
rpkyrych, line 91: The test isn't valid at all.

1. You set the new username
2. You set "ROLE_USER"
3. You invoke the tested method
4. You validate that the user has the "ROLE_USER" but you did that in step 2.

How the tested method's result is compared?

src/test/java/space/obminyashka/items_exchange/service/UserServiceIntegrationTest.java
rpkyrych: Could you please update it here rather than move it up?

rpkyrych (changes requested):

src/test/java/space/obminyashka/items_exchange/end2end/UserFlowTest.java
```diff
@@ -168,15 +168,14 @@ void selfDeleteRequest_WhenDataCorrect_Successfully() throws Exception {
     @WithMockUser(username = "deletedUser", roles = "SELF_REMOVING")
     @DataSet(value = {"database_init.yml", "user/deleted_user_init.yml"})
     @ExpectedDataSet(value = "user/deleted_user_restore_expected.yml", orderBy = {"created", "name"},
-            ignoreCols = {"password", "lastOnlineTime", "updated"})
+            ignoreCols = {"password", "lastOnlineTime", "updated", "status"})
     void makeAccountActiveAgain_WhenDataCorrect_Successfully() throws Exception {
         UserDeleteFlowDto userDeleteFlowDto = new UserDeleteFlowDto(CORRECT_OLD_PASSWORD, CORRECT_OLD_PASSWORD);
```
rpkyrych, line 173:
```suggestion
        var userDeleteFlowDto = new UserDeleteFlowDto(CORRECT_OLD_PASSWORD, CORRECT_OLD_PASSWORD);
```

rpkyrych (approved):

IllyaZuew (approved):

PR #587, opened 2023-05-30
[OBMIN-403] Optimization of SLQ 

ImAndVlad:
Optimization of SLQ requests to Get User's Advertisements API:
- added request getUserAvatarByUsername() where we get user avatar
- added new buildAdvertisementTitleDto() for findAllByUsername()

### DIFF
--- a/src/main/java/space/obminyashka/items_exchange/controller/UserController.java
+++ b/src/main/java/space/obminyashka/items_exchange/controller/UserController.java
@@ -83,7 +83,9 @@ public class UserController {
             @ApiResponse(responseCode = "404", description = "NOT FOUND")})
     @ResponseStatus(HttpStatus.OK)
     public List<AdvertisementTitleDto> getCreatedAdvertisements(@Parameter(hidden = true) Authentication authentication) {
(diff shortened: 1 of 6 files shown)


src/main/java/space/obminyashka/items_exchange/controller/UserController.java
```diff
@@ -83,7 +83,9 @@ public String updateUserInfo(@Valid @RequestBody UserUpdateDto userUpdateDto, @P
             @ApiResponse(responseCode = "404", description = "NOT FOUND")})
     @ResponseStatus(HttpStatus.OK)
     public List<AdvertisementTitleDto> getCreatedAdvertisements(@Parameter(hidden = true) Authentication authentication) {
-        return advService.findAllByUsername(authentication.getName());
+        var username = authentication.getName();
+        var avatarImage = userService.getAvatarByUsername(username);
```
rpkyrych, line 87: That is a mistake and has to be fixed.
The advertisement has to have its title image NOT the user's one.

rpkyrych (changes requested):
Your code broke the existing flow. Please fix it

PR #652, opened 2023-08-07
OBMIN-470 There is an incorrect error message when the parameter of endpoint has "null" value 

IllyaZuew:
Changes:
- added new message for null parameter validations


src/main/resources/messages.properties
```diff
@@ -5,6 +5,7 @@
 invalid.not-positive.id=ID value has to be positive
 invalid.not-empty=${validatedValue} has to be not empty
 invalid.not-null=${validatedValue} has to be not null
+invalid.not-null-phone=phone number should not be null
```
rpkyrych, line 8: You should keep ${validatedValue} for that message as well. That is our agreement with FE guys

src/main/resources/messages_ua.properties
```diff
@@ -5,6 +5,7 @@
 invalid.not-positive.id=\u0417\u043D\u0430\u0447\u0435\u043D\u043D\u044F ID \u043C\u0430\u0454 \u0431\u0443\u0442\u0438 \u043F\u043E\u0437\u0438\u0442\u0438\u0432\u043D\u0438\u043C
 invalid.not-empty=${validatedValue} \u043C\u0430\u0454 \u0431\u0443\u0442\u0438 \u043D\u0435 \u043F\u0443\u0441\u0442\u0438\u043C
 invalid.not-null=${validatedValue} \u043C\u0430\u0454 \u0431\u0443\u0442\u0438 \u043D\u0435 null
+invalid.not-null-phone=\u043D\u043E\u043C\u0435\u0440 \u0442\u0435\u043B\u0435\u0444\u043E\u043D\u0443 \u043C\u0430\u0454 \u0431\u0443\u0442\u0438 \u043D\u0435 null
```
rpkyrych, line 8: ${validatedValue} should be here either

